PR #5, opened 2016-10-08
fix compatible WordPress version in readme

bolint:
I tested the plugin on a WordPress 4.6.1 site. It worked perfectly. I'm eager to see the updated plugin but meanwhile the update of the compatible WordPress version in the readme (and on the WP's plugin page) is nice, and increase users' trust in the plugin.


const-cloudinary:
Thank you for contribution! Unfortunately, this PR is a bit outdated.

Issue #3, opened 2016-06-21
Deprecated media_buttons_context hook conflicts with other plugins


monecchi:
It seems the `media_buttons_context` hook has been deprecated . [WordPress developer docs](https://developer.wordpress.org/reference/hooks/media_buttons_context/) says to use `media_buttons` action instead. This also seems to trigger a bug if other plugins use the same deprecated hook, making the other plugins custom media buttons inserted with the `media_buttons_context` hook to be repeated several times right beside the **"Cloudinary Upload/Insert"** button.

My theme also uses the `media_buttons_context` hook to place a media button, and that was causing the button to be repeated beside the  **"Cloudinary Upload/Insert"** button too. Although I've managed to solve part of the issue by fixing my theme's code, I also use another plugin that uses the same hook, and then the bug was rising again.

So digging the cloudinary plugin code I got to manually change it to the following bellow, what successfully solved all the issues mentioned before:

Code from **cloudinary.php** [L. 494 - 502](https://github.com/cloudinary/cloudinary_wordpress/blob/master/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/cloudinary.php#L494-L502)

```
function media_cloudinary($editor_id = 'content') {
  $context = __('Cloudinary Upload/Insert');
  $xdmremote = $this->prepare_cloudinary_media_lib_url("wp_post");
  if (!$xdmremote) return ""; 

  echo $this->init_media_lib_integration($xdmremote, false) .
       '<a href="#" class="cloudinary_add_media" id="' . esc_attr( $editor_id ) . '-add_media" ' .
       'title="' . esc_attr__( 'Add Media from Cloudinary' ) . '">' . $context . '</a><span class="cloudinary_message"></span>';
  add_action('media_buttons', 'media_cloudinary');
}
```

P.S I'm not a developer, so I can't guarantee the fix above is the right way to make it, but it seems to comply with WordPress most recent version.

taragano:
@monecchi - Thank you for your feedback. While there isn't a specific ETA for this, it's indeed on our road-map to update the WordPress plugin, including the referred issue.

PR #6, opened 2016-10-08
fix button compatibility

bolint:
Removed the problematic filter.
There's no need to add another `add_filter` as suggested in the issue #3.


const-cloudinary:
Fixes #6

PR #9, opened 2017-06-06
add folder option

yonishiy:


const-cloudinary:
Thank you for contribution. This fix is resolved in #20.

PR #14, opened 2017-11-22
Changing returned URL from http to https

marissa-masangcay:
Changing the returned URL "$result["url"] => $result["secure_url"]" to return https instead of http


tocker:
Merged manually.

PR #17, opened 2017-12-14
fixed issue in cloudinary.php file when it loads files from include path instead of the local directory

garikmmm:


tocker:
@garikmmm thanks for the pull request!

PR #19, opened 2018-02-14
Fix PHP errors( notices) in DEBUG mode

const-cloudinary:
Fix `build_resize_url` handling of `full` and unsupported size strings
Declare variables that might be used before declaration
Add validation of undefined variables
Comment out unreached code
Improve code readability


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/cloudinary.php
```diff
@@ -288,49 +299,83 @@ function fix_url($url, $post_id) {
...
-        if ($resized) {
-          list ($dst_x, $dst_y, $src_x, $src_y, $dst_w, $dst_h, $src_w, $src_h) = $resized;
-          $transformation = "c_crop,h_$src_h,w_$src_w,x_$src_x,y_$src_y/";
+    // Check if this is a Cloudinary URL
+    if (!preg_match('#(.*?)/(v[0-9]+/.*)$#', $url, $matches)) {
+      return false;
+    }
+
+    if (!$size) {
+      return array($url, $metadata["width"], $metadata["height"], false);
```
const-cloudinary: @konforti , looks like the last argument(boolean for whether the image is an intermediate size) should be false, we had true. Can you please confirm?

PR #20, opened 2018-02-18
Add cloudinary_php to wordpress plugin

const-cloudinary:
Added "src" dir of cloudinary_php.
Added $USER_PLATFORM
Created aliases for backwards compatibility and marked them as deprecated


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/cloudinary.php
```diff
@@ -8,14 +8,14 @@
...
+
 define('cloudinary_VERSION', '1.1.7');
 define('cloudinary_PLUGIN_URL', plugin_dir_url( __FILE__ ));
-require "cloudinary_api.php" ;
-require "api.php" ;
-require "uploader.php" ;
-
 define ('CLOUDINARY_BASE_URL', "https://cloudinary.com");
 define ('CLOUDINARY_UNIQUE_ID', "cloudinary-image-management-and-manipulation-in-the-cloud-cdn");
+define ('CLOUDINARY_USER_PLATFORM_TEMPLATE', "CloudinaryWordpress/%s (Wordpress %s; PHP %s)");
```
tocker: Replace WordPress instead of Wordpress (capital P) in all sources

PR #26, opened 2018-06-20
Add CLI command for syncing media

tormjens:
Syncing lots of media to Cloudinary via the bulk action is less than optimal. For most setups it will time out if you try to add a large number of images. And boy does it take a long time to do it a few images at a time.

Since I am lazy and would rather write something that will do the job for me I wrote a WP CLI command for running the upload process via cli.

To start uploading all your media, you simply use (if you have WP CLI installed globally):
```
wp cloudinary-import
```


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/lib/Cli.php
```diff
@@ -0,0 +1,41 @@
...
+            }
+        }
+
+        public function uploadAttachment($attachment)
+        {
+            return $this->plugin->upload_to_cloudinary($attachment->ID, true);
+        }
+
+    }
+}
```
QuentinMichinot: Forget ligne at the end of code ?!

const-cloudinary:
Merged manually, see #27.

Thank you for contribution!

stefanomonteiro:
Is there any command to migrate away?

shirlymanor:
Yes, You can go to your media and choose to detach. You can do it for each image or by choosing all the images and choose to migrate away.

stefanomonteiro:
Using a CLI command?

shirlymanor:
Currently, we are not supporting CLI command. only from the UI.

PR #40, opened 2020-01-30
Prevent assigning error to `$this->usage`

kienstra:
There was a fatal error reported on activation:
>An error of type E_ERROR was caused in line 340 of the file /path/to/
wp-content/plugins/cloudinary_wordpress_v2/php/sync/class-push-sync.php. Error 
message: Uncaught Error: Cannot use object of type WP_Error as array in 
/path/to/wp-content/plugins/cloudinary_wordpress_v2/php/sync/
class-push-sync.php:340

https://github.com/cloudinary/cloudinary_wordpress/blob/9b10a9815b924fb64a6eebd1d8f8db6de2bd35a1/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-push-sync.php#L340

I couldn't reproduce that, with the same versions as reported in the bug:
PHP 7.4
WP Core 5.3.2
Foodoholic theme

But this PR addresses one possible cause of the error.

It might be possible that `$this->plugin->components['connect']->usage` is a `WP_Error`, causing this fatal error.

It looks like the code anticipates `$stats` being a `WP_Error`:
https://github.com/cloudinary/cloudinary_wordpress/blob/9b10a9815b924fb64a6eebd1d8f8db6de2bd35a1/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-connect.php#L285

If so, it will be assigned to `$this->plugin->components['connect']->usage`:

https://github.com/cloudinary/cloudinary_wordpress/blob/9b10a9815b924fb64a6eebd1d8f8db6de2bd35a1/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-connect.php#L291

...possibly causing the error the user reported.

Still, I can't reproduce the error that was reported. So it's only a guess that this might fix it.

### DIFF
--- a/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-connect.php
+++ b/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-connect.php
@@ -279,10 +279,14 @@ class Connect implements Config, Setup, Notice {
 
 			$this->api = new Connect\Api( $this, $this->plugin->version );
 			$stats     = get_transient( '_cloudinary_usage' );
+			if ( is_wp_error( $stats ) ) {
+				return;
+			}
+
 			if ( empty( $stats ) ) {
 				// Get users plan.
 				$stats = $this->plugin->components['connect']->api->usage();
-				if ( ! is_wp_error( $stats ) && ! empty( $stats['media_limits'] ) ) {
+				if ( ! empty( $stats['media_limits'] ) ) {
 					$stats['max_image_size'] = $stats['media_limits']['image_max_size_bytes'];
 					$stats['max_video_size'] = $stats['media_limits']['video_max_size_bytes'];
 					set_transient( '_cloudinary_usage', $stats, HOUR_IN_SECONDS );
--- a/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-connect.php
+++ b/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-connect.php
@@ -279,7 +279,6 @@ class Connect implements Config, Setup, Notice {
 
 			$this->api = new Connect\Api( $this, $this->plugin->version );
 			$stats     = get_transient( '_cloudinary_usage' );
-
 			if ( empty( $stats ) ) {
 				// Get users plan.
 				$stats = $this->plugin->components['connect']->api->usage();
--- a/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-connect.php
+++ b/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-connect.php
@@ -286,7 +286,7 @@ class Connect implements Config, Setup, Notice {
 			if ( empty( $stats ) ) {
 				// Get users plan.
 				$stats = $this->plugin->components['connect']->api->usage();
-				if ( ! empty( $stats['media_limits'] ) ) {
+				if ( ! is_wp_error( $stats ) && ! empty( $stats['media_limits'] ) ) {
 					$stats['max_image_size'] = $stats['media_limits']['image_max_size_bytes'];
 					$stats['max_video_size'] = $stats['media_limits']['video_max_size_bytes'];
 					set_transient( '_cloudinary_usage', $stats, HOUR_IN_SECONDS );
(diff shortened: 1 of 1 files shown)


kienstra:
**Request For Review**

Hi @dugajean,
Thanks a lot for reviewing #39. Could you please review this?

Thanks, Dugi!

kienstra:
Sorry, not planning on making any more commits before review 😄

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-connect.php
```diff
@@ -282,7 +282,12 @@ public function setup() {
 			if ( empty( $stats ) ) {
 				// Get users plan.
 				$stats = $this->plugin->components['connect']->api->usage();
-				if ( ! is_wp_error( $stats ) && ! empty( $stats['media_limits'] ) ) {
+
+				if ( is_wp_error( $stats ) ) {
+					return;
+				}
+
+				if ( ! empty( $stats['media_limits'] ) ) {
```
dugajean: I think it might be a better idea to set some defaults before returning if there's a `WP_Error` since there are 5 uses of its keys around the plugin, which might cause `undefined index` errors.

- See `class-media.php` at line 1079-1080
- See `class-push-sync.php` at line 340-341
- See `class-upload-sync.php` at line 120

What do you think @kienstra ?

kienstra: Uf, good point

kienstra:
**Closing**

This PR is only a guess at one way the fatal error reported could have happened. Since I can't reproduce the issue, I don't think it makes sense to continue. 

If there really was a WP_Error from calling the API:

https://github.com/cloudinary/cloudinary_wordpress/blob/9b10a9815b924fb64a6eebd1d8f8db6de2bd35a1/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/connect/class-api.php#L581

...then there would need to be better UI handling for this. The user would probably need to know that there was some kind of error in order to try again.

Simply settings `'max_image_size'` to a default like 0 or PHP_INT_MAX doesn't seem fair, as it might not have been their fault.

I don't think we can 'fix' something that's not reproducible.

const-cloudinary:
@kienstra , I can reproduce this, we have an environment that fails with this issue.

kienstra:
@const-cloudinary,
Thanks, are there any other active plugins, other than Cloudinary? And what theme is active?

The stack trace here unfortunately doesn't point to where the WP_Error could be coming from.

const-cloudinary:
@kienstra, I would start with invalid CLOUDINARY_URL, when "usage" api returns 40x/50x response or something like that.

const-cloudinary:
@kienstra I printed the error. it says:
disabled account in /home/consta01/public_html/wp-content/plugins/cloudinary_wordpress_v2 5/php/sync/class-push-sync.php...

kienstra:
Hi @const-cloudinary,
Thanks for your help.

>@kienstra I printed the error. it says:
disabled account in /home/consta01/public_html/wp-content/plugins/cloudinary_wordpress_v2 5/php/sync/class-push-sync.php...

Would you be able to paste the rest of the error, especially the line number?

const-cloudinary:
@kienstra It's the same line of the "Cannot use object of type WP_Error as array" . WP_Error contains this message. 

You can reproduce it by disabling your Cloudinary account, clicking "Cancel my account" on this page:
https://cloudinary.com/console/settings/account

kienstra:
Hi @const-cloudinary,
Thanks for helping with the step to reproduce this.

There didn't seem to be a PHP error after canceling my Cloudinary account:

activating-cloud

I also tried `wp transient delete _cloudinary_usage`, and then activating the plugin.

const-cloudinary:
@kienstra Can you please provide me some code that I can run on my environment (add some logging entries/write notice message on the screen, etc), so I can understand what is exactly happening in my environment?

kienstra:
Hi @const-cloudinary,
Thanks, do you have `xdebug` enabled? That's probably going to be easier than `var_dump()` or `print()` statements.

Also, this debugging tool has helped me a lot: https://github.com/nette/tracy

const-cloudinary:
@kienstra This bug is reproducible on a shared hosting website I own, it will be problematic to install xdebug there.

kienstra:
Hi @const-cloudinary,
OK, thanks.

Is that a production site? If not, would it be possible to deploy this PR of the plugin to it, to see if it fixes it?

If it's a production site, would it be possible to create a staging 'clone' of it?

Thanks!

PR #42, opened 2020-02-06
Cloudinary video player does not conform to transformation string set in the settings

dugajean:
After a lot of research on this bug and a lot of failed attempts, I'm finally happy with my solution. 

As Ryan has described, steps to reproduce:

- In /wp-admin > Cloudinary > Global Transformations > Video Settings, select ‘Cloudinary Player'
- Enter h100,w100 as the ‘Video Transformation String’
- Click ‘Save Settings’
- Create a new post in the block editor
- Add a video block
- Click ‘Media Library’
- Select a video that has been synced with Cloudinary, or upload a video and ensure it’s synced
- Click ‘Select'
- Click ‘Preview’ to see the front-end
- Expected: because there’s a global transformation of h100,w100, the video should be 100 x 100
- Actual: the video is its normal size

The Cloudinary video player JS library has a ton of options that can be added when instantiating it (including `width` and `height`), but simply parsing `w_100` and `h_100` into an options object and passing it didn't seem like a very scalable solution as there are a lot more transformations that could be applied.

I realized that transformations could be directly added to the URL in their short form (ie. w_100), so parsing the transformations is no longer needed.

Example:
> http://res.cloudinary.com/myuser/video/upload/vid.mp4

Turns into:
> http://res.cloudinary.com/myuser/video/upload/w_100,h_100/vid.mp4

I generate a small amount of JS to switch the URL for each Cloudinary video on the page and apply the global transformations accordingly.

Result:

Screen Shot 2020-02-06 at 21 09 27

### DIFF
--- a/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-video.php
+++ b/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-video.php
@@ -391,17 +391,6 @@ window.addEventListener( 'load', function() {
 	}
 
 	/**
-	 * Wraps a JS string in a window.onload callback.
-	 *
-	 * @param string $code
-	 * 
-	 * @return string
-	 */
-	protected function window_onload_wrapper( $code ) {
-		return $code ? "\nwindow.addEventListener( 'load', function () { {$code} });" : '';
-	}
-
-	/**
 	 * Enqueue BLock Assets
 	 */
 	public function enqueue_block_assets() {
--- a/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-video.php
+++ b/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-video.php
@@ -344,7 +344,7 @@ class Video {
 				}
 
 				$code[] = sprintf(
-					'var video_%1$s = cld.videoPlayer( "cloudinary-video-%1$s", %2$s );',
+					'var video%1$s = cld.videoPlayer( "cloudinary-video-%1$s", %2$s );',
 					$instance,
 					wp_json_encode( $config )
 				);
@@ -352,10 +352,13 @@ class Video {
 				if ( isset( $this->config['video_freeform'] ) ) {
 					$code[] = sprintf(
 						'window.onload = function () {
-							var videoContainer = document.getElementById( video_%s.videojs.id_ );
-							videoContainer.style.overflow = "hidden";
-							var videoElement = videoContainer.getElementsByTagName( "video" )[0];
-							videoElement.src = videoElement.src.replace( "upload/", "upload/%s/" );
+							var videoContainer%1$s = document.getElementById( video%s.videojs.id_ );
+							var videoElement%1$s = videoContainer%1$s.getElementsByTagName( "video" )[0];
+							videoElement%1$s.src = videoElement%1$s.src.replace( "upload/", "upload/%2$s/" );
+
+							if ( videoElement%1$s.width < 320 ) {
+								video%1$s.controls(false);
+							}
 						};', 
 						$instance, 
 						$this->config['video_freeform']
--- a/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-video.php
+++ b/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-video.php
@@ -353,7 +353,7 @@ class Video {
 
 			ob_start();
 			?>
-var cldVideos = <?php echo wp_json_encode( $cld_videos ); ?>;
+			var cldVideos = <?php echo wp_json_encode( $cld_videos ); ?>;
 
 for ( var videoInstance in cldVideos ) {
 	var config = cldVideos[ videoInstance ];
--- a/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-video.php
+++ b/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-video.php
@@ -378,8 +378,9 @@ window.addEventListener( 'load', function() {
 		}
 	}
 } );
-<?php endif ?>
-			<?php
+			<?php 
+			endif; 
+			
 			$script = ob_get_clean();
 
 			wp_add_inline_script(
--- a/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-video.php
+++ b/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-video.php
@@ -356,24 +356,23 @@ class Video {
 			var cldVideos = <?php echo wp_json_encode( $cld_videos ); ?>;
 
 for ( var videoInstance in cldVideos ) {
-	var config = cldVideos[ videoInstance ];
-	var id = 'cloudinary-video-' + videoInstance;
-	cld.videoPlayer( id, config );
+	var cldConfig = cldVideos[ videoInstance ];
+	var cldId = 'cloudinary-video-' + videoInstance;
+	cld.videoPlayer( cldId, cldConfig );
 }
 
 <?php if ( $this->config['video_freeform'] ): ?>
 window.addEventListener( 'load', function() {
-	for ( const instance in cldVideos ) {
-		var config = cldVideos[ instance ];
-		var id = 'cloudinary-video-' + instance;
-		var videoContainer = document.getElementById( id );
+	for ( var videoInstance in cldVideos ) {
+		var cldId = 'cloudinary-video-' + videoInstance;
+		var videoContainer = document.getElementById( cldId );
 		var videoElement = videoContainer.getElementsByTagName( 'video' );
 
 		if ( videoElement.length === 1 ) {
 			videoElement = videoElement[0];
 			videoElement.src = videoElement.src.replace( 
 				'upload/', 
-				'upload/<?php echo esc_js( $this->config['video_freeform'] ) ?>' 
+				'upload/<?php echo esc_js( $this->config['video_freeform'] ) ?>/' 
 			);
 		}
 	}
--- a/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-video.php
+++ b/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-video.php
@@ -320,7 +320,7 @@ class Video {
 
 		if ( $this->player_enabled() && ! empty( $this->attachments ) ) {
 
-			$code = $onload_code = array();
+			$cld_videos = array();
 			foreach ( $this->attachments as $instance => $video ) {
 				// @todo - ping the URL to ensure it has transformation available, else update an eager.
 				$cloudinary_id = $this->media->get_public_id( $video['id'] );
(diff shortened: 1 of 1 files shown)


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-video.php
```diff
@@ -338,19 +338,62 @@ public function print_video_scripts() {
...
+				if ( empty( $config['size'] ) && ! isset( $this->config['video_freeform'] ) ) {
 					$config['fluid'] = true;
 				}
-				$code[] = 'cld.videoPlayer(\'cloudinary-video-' . $instance . '\', ' . wp_json_encode( $config ) . ');';
+
+				$code[] = sprintf(
+					"var video%1\$s = cld.videoPlayer( \"cloudinary-video-%1\$s\", %2\$s );\n",
+					$instance,
+					wp_json_encode( $config )
+				);
```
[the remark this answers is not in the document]
dugajean: This does indeed fix the code repetition in the JS. I'll implement it this way. Thanks!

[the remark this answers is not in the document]
dugajean: @kienstra I tried implementing this, but it's a little troublesome as initializing the player onload is a little too late. Swapping the logic of the init outside the event causes weird behavior (only the last video instance receives the changes).

So I took your other suggestions for in this review for now, and I'll potentially revisit this in the future.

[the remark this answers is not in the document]
kienstra: Hi @dugajean,
Ah, good point.

For the script, how about something like:

```js

			ob_start();
			?>
			const cldVideos = <?php echo wp_json_encode( $cld_videos ); ?>;

			for ( const videoInstance in cldVideos ) {
				let config = cldVideos[ videoInstance ];
				let id = 'cloudinary-video-' + videoInstance;
				cld.videoPlayer( id, config );
			}

			window.addEventListener( 'load', function() {
				for ( const instance in cldVideos ) {
					let config = cldVideos[ instance ];
					let id = 'cloudinary-video-' + instance;
					let videoContainer = document.getElementById( id );
					let videoElement = videoContainer.getElementsByTagName( 'video' );

					if ( videoElement.length === 1 ) {
						videoElement = videoElement[0];
						videoElement.src = videoElement.src.replace( 'upload/', 'upload/' + instance + '/' );
					}
				}
			} );
			<?php
			$script = ob_get_clean();
```

This omits:

```js
if ( videoElement.width < 320 ) {
	video.controls(false);
}
```

You've looked at this more than me, so maybe that part is necessary.

But it looks like there already is a way to globally disable the controls:

show-controls

This seems like it should be handled by the user or the Cloudinary API, not this script.

But you've looked at this more than I have 😄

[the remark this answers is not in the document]
dugajean: Ah, I didn't explain that. I forcefully turnoff the controls because they go out of bounds when the player is less than 320px. This way we can ensure that we wont get a broken UI.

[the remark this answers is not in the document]
dugajean: I'll also test your code generation changes. Thanks for the input on that 😄

dugajean:
@kienstra I'll do that then! 👍

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-video.php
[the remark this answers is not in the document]
kienstra: OK, thanks for elaborating.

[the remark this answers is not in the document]
kienstra: Do you think this plugin should be disabling controls, when the user can presumably do that via that screenshot above? Or maybe here also:

playback-controls

dugajean:
@kienstra Implemented your suggestion from above. Works flawlessly!

Also kept the controls check removed as you noted above and let it up to the user to remove the controls if they see it as necessary.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-video.php
```diff
@@ -343,43 +343,49 @@ public function print_video_scripts() {
...
-					implode( $code ) . $this->window_onload_wrapper( implode( $onload_code ) ) 
-				);
+
+			if ( empty( $cld_videos ) ) {
+				return;
 			}
+
+			ob_start();
+			?>
+var cldVideos = <?php echo wp_json_encode( $cld_videos ); ?>;
```
kienstra: Nitpick: could this be indented at the same level as the `?>` above? Maybe there is a reason, it just looks a little strange to me.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-video.php
```diff
@@ -343,43 +343,49 @@ public function print_video_scripts() {
...
+
+for ( var videoInstance in cldVideos ) {
+	var config = cldVideos[ videoInstance ];
+	var id = 'cloudinary-video-' + videoInstance;
```
kienstra: Could `config` and `id` use `let`, so they're in the scope of this `for` block? Otherwise, they're in the global scope, I think.

dugajean: Is it safe to use `const/let` without transpilation taking place? Some older versions of browsers may not work well with non-ES5 JS.

kienstra: Hm...

kienstra: Good point, there's [95% support](https://caniuse.com/#feat=let), so maybe just `var cldConfig` and `var cldId` would be better. 

Or something like that.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-video.php
```diff
@@ -338,16 +338,55 @@ public function print_video_scripts() {
...
+
+			if ( empty( $cld_videos ) ) {
+				return;
 			}
+
+			ob_start();
+			?>
+			var cldVideos = <?php echo wp_json_encode( $cld_videos ); ?>;
+
+for ( var videoInstance in cldVideos ) {
```
kienstra: Ah, I meant indenting this entire script to the same level. What do you think?

Like:

```php
	public function print_video_scripts() {

		if ( $this->player_enabled() && ! empty( $this->attachments ) ) {

			$cld_videos = array();
			foreach ( $this->attachments as $instance => $video ) {
				// @todo - ping the URL to ensure it has transformation available, else update an eager.
				$cloudinary_id = $this->media->get_public_id( $video['id'] );
				$default       = array(
					'publicId'    => $cloudinary_id,
					'sourceTypes' => array( $video['format'] ), // @todo Make this based on eager items as mentioned above.
					'controls'    => 'on' === $this->config['video_controls'] ? true : false,
					'autoplay'    => 'off' !== $this->config['video_autoplay_mode'] ? true : false,
					'loop'        => 'on' === $this->config['video_loop'] ? true : false,
				);

				$valid_autoplay_modes = [ 'never', 'always', 'on-scroll' ];
				if ( $default['autoplay'] && in_array( $this->config['video_autoplay_mode'], $valid_autoplay_modes, true ) ) {
					$default['autoplayMode'] = $this->config['video_autoplay_mode'];
				}

				$config = wp_parse_args( $video['args'], $default );

				if ( empty( $config['size'] ) && ! isset( $this->config['video_freeform'] ) ) {
					$config['fluid'] = true;
				}

				$cld_videos[ $instance ] = $config;

			}

			if ( empty( $cld_videos ) ) {
				return;
			}

			ob_start();
			?>
			var cldVideos = <?php echo wp_json_encode( $cld_videos ); ?>;
			
			for ( var videoInstance in cldVideos ) {
				var config = cldVideos[ videoInstance ];
				var id = 'cloudinary-video-' + videoInstance;
				cld.videoPlayer( id, config );
			}
			
			<?php if ( $this->config['video_freeform'] ): ?>
			window.addEventListener( 'load', function() {
				for ( const instance in cldVideos ) {
					var config = cldVideos[ instance ];
					var id = 'cloudinary-video-' + instance;
					var videoContainer = document.getElementById( id );
					var videoElement = videoContainer.getElementsByTagName( 'video' );
			
					if ( videoElement.length === 1 ) {
						videoElement = videoElement[0];
						videoElement.src = videoElement.src.replace( 
							'upload/', 
							'upload/<?php echo esc_js( $this->config['video_freeform'] ) ?>' 
						);
					}
				}
			} );
			<?php 
			endif; 
			
			$script = ob_get_clean();

			wp_add_inline_script(
				'cld-player',
				$script
			);
		}
	}
```

kienstra: Ah, I meant indenting this entire script to the same level. What do you think?

Like:

```php
	public function print_video_scripts() {

		if ( $this->player_enabled() && ! empty( $this->attachments ) ) {

			$cld_videos = array();
			foreach ( $this->attachments as $instance => $video ) {
				// @todo - ping the URL to ensure it has transformation available, else update an eager.
				$cloudinary_id = $this->media->get_public_id( $video['id'] );
				$default       = array(
					'publicId'    => $cloudinary_id,
					'sourceTypes' => array( $video['format'] ), // @todo Make this based on eager items as mentioned above.
					'controls'    => 'on' === $this->config['video_controls'] ? true : false,
					'autoplay'    => 'off' !== $this->config['video_autoplay_mode'] ? true : false,
					'loop'        => 'on' === $this->config['video_loop'] ? true : false,
				);

				$valid_autoplay_modes = [ 'never', 'always', 'on-scroll' ];
				if ( $default['autoplay'] && in_array( $this->config['video_autoplay_mode'], $valid_autoplay_modes, true ) ) {
					$default['autoplayMode'] = $this->config['video_autoplay_mode'];
				}

				$config = wp_parse_args( $video['args'], $default );

				if ( empty( $config['size'] ) && ! isset( $this->config['video_freeform'] ) ) {
					$config['fluid'] = true;
				}

				$cld_videos[ $instance ] = $config;

			}

			if ( empty( $cld_videos ) ) {
				return;
			}

			ob_start();
			?>
			var cldVideos = <?php echo wp_json_encode( $cld_videos ); ?>;
			
			for ( var videoInstance in cldVideos ) {
				var config = cldVideos[ videoInstance ];
				var id = 'cloudinary-video-' + videoInstance;
				cld.videoPlayer( id, config );
			}
			
			<?php if ( $this->config['video_freeform'] ): ?>
			window.addEventListener( 'load', function() {
				for ( const instance in cldVideos ) {
					var config = cldVideos[ instance ];
					var id = 'cloudinary-video-' + instance;
					var videoContainer = document.getElementById( id );
					var videoElement = videoContainer.getElementsByTagName( 'video' );
			
					if ( videoElement.length === 1 ) {
						videoElement = videoElement[0];
						videoElement.src = videoElement.src.replace( 
							'upload/', 
							'upload/<?php echo esc_js( $this->config['video_freeform'] ) ?>' 
						);
					}
				}
			} );
			<?php 
			endif; 
			
			$script = ob_get_clean();

			wp_add_inline_script(
				'cld-player',
				$script
			);
		}
	}
```

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-video.php
dugajean: Should be good now.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-video.php
dugajean: The outputted JS looks a weirdly indented that way. But yeah, I don't really mind that as nobody's going to go through the source page anyway 😄

dugajean:
@kienstra Should be all good now! 🎉

kienstra:
Feel free to merge this when you'd like.

kienstra:
I'm merging this, but if there's something you'd like to change, feel free to open another PR.

PR #50, opened 2020-03-31
Version2.0.0 release

asisayag2:
Updated the readme + screenshots


asisayag2:
@DavidCramer , I created this PR for preparing for the release to the Market place. 
I updated the readme + screenshots. 
I also removed the updater folder as this is not part of the release (maybe we can add under some utils folder?)
Please review and let us know if something in the readme needs to be changed. if not, please merge

DavidCramer:
> I also removed the updater folder as this is not part of the release (maybe we can add under some utils folder?)

@asisayag2 - that may be a good idea if you like, but it's not necessary.

PR #73, opened 2020-04-23
Add breakpoint flag to convert_url()

pereirinha:
This will allow applying the default transformations when calling the_post_thumbnail().

### DIFF
--- a/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
+++ b/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
@@ -463,7 +463,7 @@ class Media implements Setup {
 	 * @return string Cloudinary URL.
 	 */
 	public function attachment_url( $url, $attachment_id ) {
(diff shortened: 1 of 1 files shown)


DavidCramer:
@dugajean - the `the_post_thumbnail()` calls the `image_downsize` filter which allows us to replace a sized URL for an image. The problem with that is that we filter this by converting the URL, which is an exact copy without global transformations, since `convert_url` was initially made to convert breakpoints by passing an array of transformations from the parent.
The result is a clean, untransformed URL.

The flag that @pereirinha added, lets us selectively tell it to apply globals transformations, with it set to false in the breakpoint context.

This however is only a partial fix, since it doesn't take into account if the featured image has it's own transformations. I have pushed a fix that uses the correct way, by converting the primary URL, and having the breakpoints part be an accessory rather than an override. 
I also renamed the `$breakpoint` flag to better explain it's purpose; `Overwrite Transformations`

dugajean:
Got it @DavidCramer, thanks for explaining. 😄

PR #76, opened 2020-04-24
Feature/CLOUD-396

DavidCramer:
Alert/Admin notices shown at 70%, 80%, and 90% (with a link to upgrade at 90%)
Screenshot 2020-04-24 at 12 28 16

Dashboard Settings page added usage summary.
Screenshot 2020-04-24 at 12 29 10


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/ui-definitions/tabs/dashboard-content.php
```diff
@@ -7,6 +7,27 @@
 
 $video_url = ''; // Left blank for when we get the final video URL.
 $usage     = $this->plugin->components['connect']->usage;
+// Different account version have usage limits.
+if ( isset( $usage['credits'] ) ) {
+	// New credit system.
+	$usage['credits']         = array(
+		'limit'        => $usage['credits']['limit'],
```
pereirinha, line 14: @DavidCramer I notice that we are checking if the `credits` key is set before we run this logic, but I wonder if that's enough. My concerns here are what about all these child keys, that might not be set. To add to it, it seems that we are getting this information from a source that we don't control. Ideally, this should be addressed on `usage_stats()` — setting some defaults —, so we can be relaxed using the logic elsewhere. Thoughts?

DavidCramer: @pereirinha - That is a good suggestion. However, the issue is that there are two different style of settings, although we could switch between the defaults.
Also, this is directly from their API, and is documented. So it shouldn't really change. If it did, we would have it break without breaking.

An ideal way, would be to check when calling the API, then output an error before actually setting the value.
I'll let this pass for now, then loop back with a clean up ticket.

DavidCramer:
@pereirinha - I altered it a little by adding in a method to fetch a specific stat. This will ensure we never reference something that doesn't exist.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-connect.php
```diff
@@ -290,23 +313,62 @@ public function setup() {
...
+			$this->usage_stats( true );
+		}
+	}
+
+	/**
+	 * Set the usage stats from the Cloudinary API.
+	 *
+	 * @param bool $refresh Flag to force a refresh.
+	 */
+	public function usage_stats( $refresh = false ) {
```
pereirinha, line 327: @DavidCramer 

It seems that this method is never called to get stalled data. Because of that, it's always making a new request to the API to get fresh data.
Can you confirm that's the case?

DavidCramer: @pereirinha - You are right! I made the `$refresh` true to test. and left it in there.
I'll make the fix for it in the morning!

DavidCramer: @pereirinha - I removed the refresh flag, and changed the text as requested by the client.

PR #84, opened 2020-05-15
Disable off option in Autoplay when necessary

dugajean:
Disable the "off" dropdown option for Autoplay if the player isn't set to Cloudinary or if Show Controls if unchecked.


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/js/src/components/settings-page.js
```diff
@@ -1,5 +1,26 @@
...
 
+	// Disable the "off" dropdown option for Autoplay if
+	// the player isn't set to Cloudinary or if Show Controls if unchecked.
+	const disableAutoplayOff = function() {
+		const player = jQuery( '#field-video_player' ).val();
+		const showControls = jQuery( '#field-video_controls' ).prop( 'checked' );
+
+		if ( player === 'cld' && showControls ) {
+			jQuery( '#field-video_autoplay_mode option[value="off"]' )
+				.attr( 'disabled', 'disabled' )
```
pereirinha, line 11: @dugajean 

I think we are duplicating the effort by both setting attr and prop. It seems they produce the [same outcome](https://api.jquery.com/prop/) in a slightly different way. I'd say it makes to most sense to use here the prop, and setting properties uses booleans, so something like this `.prop( 'disabled', true );`.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/js/src/components/settings-page.js
```diff
@@ -1,5 +1,26 @@
...
+				.prop( 'disabled', 'disabled' );
+		} else {
+			jQuery( '#field-video_autoplay_mode option[value="off"]' )
+				.removeAttr( 'disabled' )
```
pereirinha, line 15: Same here.

dugajean:
@pereirinha I addressed all your concerns about the logic and feel like this is ready to go now.

PR #94, opened 2020-06-04
fixed regex according to API Environment var

fvsqr:
Regex for checking API Environment variable is not allowing dashes in the API Secret part.
Furthermore, the cloud name could not contain numbers. As the cloud name seems to be a hex value, the regex has been modified to check for hex


dugajean:
Great change, thanks! This needs to go into the `develop` branch however. Can you please adjust the PR?

dugajean:
@fvsqr Also, cloud names can be alphanumeric and iclude hyphens as well. Could you please correct that too? Thanks!

fvsqr:
@dugajean PR moved to `develop` branch and added modified pattern for correct cloud name check in #98

PR #98, opened 2020-06-05
fixed regex according to API Environment var

fvsqr:
Regex for checking API Environment variable is not allowing dashes in the API Secret part.
Furthermore, the cloud name could not contain numbers. As the cloud name seems to be a hex value, the regex has been modified to check for hex


fvsqr:
regex allows alphanumerics and hyphens in the cloud name part.

DavidCramer:
Addressed in another update. 
Many thanks @fvsqr for your input. You were not sidelined.

PR #105, opened 2020-06-23
Feature/public id sync

DavidCramer:
Allows for being able to alter the public_id for uploading as well as correcting the folder changing sync.

**Testing notes.**

This allows for changing the public ID of assets.
Below is an example of using the filter `cloudinary_upload_options` to use the assets slug as it's `public_id` 
Add the snippet in the `functions.php` in your theme.
go to the media library page. If all assets slug is the same as the ID (it should be), then the cloud should be green.

Edit an asset, and change the slug (you may need to select Screen Options, and enable it).
Update and go back to the media library. The image should now be in the syncing state (orange cloud).
After a moment, it should be green and the public ID should be the same as the slug.

```php
	add_filter( 'cloudinary_upload_options', function ( $options, $attachment_post ) {

		/**
		 * Structure of options as follows:
		 *
		 * $options = array(
		 *    'unique_filename' => false, // Allow for auto filenames. See Cloudinary docs.
		 *    'resource_type'   => $resource_type, // Type can be 'image' or 'video' or custom.
		 *    'public_id'       => $public_id, // Changing this will change the public ID as stored in Cloudinary.
		 *    'context'         => array(
		 *        'caption' => esc_attr( $post->post_title ), // Set the caption data in Cloudinary.
		 *        'alt'     => $post->_wp_attachment_image_alt, // set the alt text of the asset in Cloudinary
		 *    ),
		 * );
		 **/

		// To change the asset public_id.
		$post_name = strtolower( $attachment_post->post_name );
		if ( $post_name !== $options['public_id'] ) {
			$options['public_id'] = $post_name;
		}

		return $options;
	}, 10, 2 );
```

### DIFF
--- a/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-push-sync.php
+++ b/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-push-sync.php
@@ -304,7 +304,7 @@ class Push_Sync {
 
 		$type = 'upload';
 		// Check for explicit (has public_id, but no breakpoints).
-		$attachment_signature = $this->plugin->components['sync']->generate_signature( $attachment->ID );
+		$attachment_signature = $this->plugin->components['sync']->get_signature( $attachment->ID );
 		if ( empty( $attachment_signature ) ) {
 			if ( ! empty( $attachment->{Sync::META_KEYS['public_id']} ) ) {
 				// Has a public id but no signature, explicit update to complete download.
@@ -312,10 +312,8 @@ class Push_Sync {
 			}
 			// fallback to upload.
 		} else {
-			// Has signature find differences.
+			// Has signature find differences and use specific sync method.
 			$required_signature = $this->plugin->components['sync']->generate_signature( $attachment->ID );
-			// Apply the generated asset signature as a default. To allow for signature extensions that may not exist.
-			$attachment_signature = wp_parse_args( $attachment_signature, $required_signature );
 			foreach ( $required_signature as $key => $signature ) {
 				if ( ( ! isset( $attachment_signature[ $key ] ) || $attachment_signature[ $key ] !== $signature ) && isset( $this->sync_types[ $key ] ) ) {
 					return $this->sync_types[ $key ];
--- a/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-push-sync.php
+++ b/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-push-sync.php
@@ -429,13 +429,13 @@ class Push_Sync {
 				}
 				$max_width = $this->plugin->components['media']->get_max_width();
 				// Add breakpoints request options.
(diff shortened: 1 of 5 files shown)


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-push-sync.php
```diff
@@ -570,6 +568,13 @@ public function push_attachments( $attachments ) {
 						$args['context'] = $upload['options']['context'];
 					}
 					$result = $this->plugin->components['connect']->api->explicit( $args );
+				} elseif ( 'rename' === $sync_type ) {
+					// Rename an asset.
+					$args   = array(
+						'from_public_id' => $this->plugin->components['media']->get_post_meta( $attachment->ID, Sync::META_KEYS['public_id'] ),
+						'to_public_id'   => $upload['public_id'],
+					);
+					$result = $this->plugin->components['connect']->api->{$upload['options']['resource_type']}( 'rename', 'POST', $args );
```
dugajean, line 577: Not sure what's going on here. This abstraction is very hard to understand without dumping out the data.

Frankly, this method is _massive_ and could use some refactoring. Very hard to understand what's going on overall.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-push-sync.php
```diff
@@ -389,30 +390,30 @@ public function prepare_upload( $post, $down_sync = false ) {
...
+			if ( false !== strpos( $public_id, '/' ) ) {
+				// Split the public_id into path and filename to allow filtering just the ID and not giving access to the path.
+				$public_id_info   = pathinfo( $public_id );
+				$public_id_folder = trailingslashit( $public_id_info['dirname'] );
+				$public_id_file   = $public_id_info['filename'];
+			} else {
+				// File is in the root of cloudinary.
+				$public_id_folder = '';
+				$public_id_file   = $public_id;
 			}
```
dugajean, line 410: Maybe we can switch to this in order to satisfy linters:

```suggestion
			// File is in the root of cloudinary.
			$public_id_folder = '';
			$public_id_file  = $public_id;
			
			if ( false !== strpos( $public_id, '/' ) ) {
				// Split the public_id into path and filename to allow filtering just the ID and not giving access to the path.
				$public_id_info   = pathinfo( $public_id );
				$public_id_folder = trailingslashit( $public_id_info['dirname'] );
				$public_id_file   = $public_id_info['filename'];
			} 
```

dugajean:
LGTM

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-push-sync.php
DavidCramer: @dugajean Nice idea. I usually do prefer to have a default, and if it's not what's expected then change.

PR #108, opened 2020-06-27
In case of a non-connected account, serve assets from WP

dugajean:
This will add a check around many action registrations and prevent them from being added if the account is not connected. The connection is checked via a new method `is_connected()` under `class-connect.php`.


DavidCramer:
@dugajean - Nice work! great solution.

DavidCramer:
@pereirinha - if we implemented this into `get_config()` in the `class-connect.php` then all those that require a connection will not be enabled and be treated as not setup.

dugajean:
I implemented Marco's suggestion and wanted to see what you guys think about this approach.

Issue #82, opened 2020-05-12
Excessive backslash stripping when editing in Gutenberg


mattleff:
With WordPress 5.4.1 and version 2.0.3 of this plugin backslashes are stripped from the content.

### Steps

1. Create new post with Gutenberg editor.
2. Enter a backslash in the content (i.e., `This is a backslash: \.`).
3. Edit the post again.

### Expected result

The content should be `This is a backslash: \.`.

### Actual result

The content is `This is a backslash: .`.

### Notes

Disabling the call to `wp_unslash()` here fixes this issue:

https://github.com/cloudinary/cloudinary_wordpress/blob/e6e69211d6f151ad8cfb6478721e5d5b3f87ecfc/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-filter.php#L523-L533

I don't know if this was necessary for prior versions of WP but it causes issues now with block data containing escape sequences (`\n`, `\r`, `\t`, etc.). I'm still determining if there is a userland workaround.

mattleff:
My current workaround:

```php
array_map(
    function ( $type ) {
        add_filter( 'rest_prepare_' . $type, function( $response, $post, $request ) {
            $context = $request->get_param( 'context' );
            if ( 'edit' === $context ) {
                $data = $response->get_data();
                $data['content']['raw'] = addslashes( $data['content']['raw'] );
                $response->set_data( $data );
            }
            return $response;
        }, 9, 3 );
    },
    get_post_types_by_support( 'editor' )
);
```

pereirinha:
I'm closing this issue, as it was resolved.
Please, feel free to reopen should you still find problems with it.

PR #115, opened 2020-07-07
Remove unslash when editing in Gutenberg

dugajean:
I tested this for potential side effects, but I can't see any case where removing slashes is necessary here.

@DavidCramer Could I be missing something here?


pereirinha:
This PR seems to be related to #82.

dugajean:
@pereirinha Oh nice, I didn't notice that issue. I'm not sure if the suggested fix is really a fix though.

PR #116, opened 2020-07-08
add upgrade handling

DavidCramer:
This is to better handle upgrading from v1 - v2+

@pereirinha - The change in `cloudinary_id` filter came into a requirement here, as this allows us to identify upgrade items and act on them without overly filtering.

### DIFF
--- a/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-upgrade.php
+++ b/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-upgrade.php
@@ -26,12 +26,22 @@ class Upgrade {
 	private $media;
 
 	/**
+	 * Holds the Sync instance.
+	 *
+	 * @since   0.1
+	 *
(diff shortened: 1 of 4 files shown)


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-upgrade.php
```diff
@@ -128,15 +139,17 @@ function ( $val ) use ( $media ) {
 		$path      = pathinfo( $public_id );
 		$public_id = strstr( $public_id, '.' . $path['extension'], true );
 		$this->media->update_post_meta( $attachment_id, Sync::META_KEYS['public_id'], $public_id );
-
+		if ( !defined( 'DOING_BULK_SYNC' ) ) {
```
dugajean, line 142:
```suggestion
		if ( ! defined( 'DOING_BULK_SYNC' ) ) {
```

PR #123, opened 2020-07-17
New sync status

pereirinha:
This PR introduces 2 new status for assets.

- downloading — for upgrades from version 1;
- syncing — for internal upgrade of the asset meta;


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-push-sync.php
```diff
@@ -391,6 +391,10 @@ public function prepare_upload( $post, $down_sync = false ) {
 				$error = sprintf( __( 'File size exceeds the maximum of %s. This media asset will be served from WordPress.', 'cloudinary' ), $max_size_hr );
 				$media->delete_post_meta( $post->ID, Sync::META_KEYS['pending'] ); // Remove Flag.
 
+				// Cleanup flags
+				delete_post_meta( $post->ID, Sync::META_KEYS['syncing'] );
+				delete_post_meta( $post->ID, Sync::META_KEYS['downloading'] );
```
dugajean, line 396: These two lines are repeated quite a bit, so maybe it'd be good to extract this into a method 😉 

```
function clear_attachment_status( $attachment_id )
```

PR #130, opened 2020-07-20
Fixes WPBakery insert error

dugajean:
Test instructions:

- Install WPBakery
- Insert Cloudinary images in a post using WPBakery editor
- Observe console and notice absence of errors


DavidCramer:
Thanks for this @dugajean - I'm curious to know, what happens if the `attach` isn't available? How is it handled?

dugajean:
@DavidCramer In such a case, it's not possible to use transformations. WPBBakery is somehow messing with these lines:

```
let state  = wp.media.editor.get().state();
let attach = state.get( 'selection' ).get( data.attachment );
```

The state with WPBakery enabled results in the `selection` state to be missing. The only option I see available here is error prevention or disabling the plugin entirely.

DavidCramer:
@dugajean - So does this only happen when you import an image from cloudinary that has a transformation?

dugajean:
@DavidCramer I've to check that out. But my guess would be that it happens every time.

pereirinha (approved):
Thanks @dugajean 
Let's move this forward.

PR #131, opened 2020-07-22
Fix syncing status

pereirinha:
The syncing status should only be available if the auto-sync feature is enabled.


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-settings-page.php
```diff
@@ -913,4 +913,19 @@ public function set_active_page( $page_slug ) {
...
+	 * @return bool
+	 */
+	public function is_auto_sync_enabled() {
+		$settings = $this->get_config();
+
+		if ( ! empty( $settings['sync_media']['auto_sync'] ) && 'on' === $settings['sync_media']['auto_sync'] ) {
+			return true;
+		}
+
+		return false;
```
dugajean, line 929: Could also be:
```suggestion
		return ! empty( $settings['sync_media']['auto_sync'] ) && 'on' === $settings['sync_media']['auto_sync'];
```

PR #132, opened 2020-07-22
Convert image and videos to blocks when switching from Classic editor to Block editor

dugajean:
Test notes:

- Have both Classic and Block editors ready for use
- Create a post using Classic Editor, insert image and/or video
- Save the post
- Switch to Gutenberg
- Image and video content should've been converted to their respective blocks rather than "Classic Block"

### DIFF
--- a/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-filter.php
+++ b/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-filter.php
@@ -585,13 +585,15 @@ class Filter {
 		$imgs = $dom->getElementsByTagName( 'img' );
 
 		foreach ($imgs as $img) {
-			// Ensure we're dealing with a proper <img> tag
-			if ( false === strpos( $img->getAttribute( 'class' ), 'wp-image-' ) ) {
+			$class_list = $img->getAttribute( 'class' );
+			
+			// Ensure we're dealing with a proper <img> tag that came from Classic Editor
+			if ( false === strpos( $class_list, 'wp-image-' ) || '' === $img->getAttribute( 'width' ) ) {
 				continue;
 			}
 
-			$img_id = preg_replace( '/[^0-9]/', '', $img->getAttribute( 'class' ) );
-			
+			list( $img_id, $size, $align, $json ) = $this->prepare_image_block_classes( $class_list );
+
 			// Prep the <img> node
 			$img_clone = $img->cloneNode( true );
 			$img_clone->removeAttribute( 'width' );
@@ -601,11 +603,11 @@ class Filter {
 			// Prep the <figure> node
 			$img_fig = $dom->createElement( 'figure' );
 			$img_fig->appendChild( $img_clone );
-			$img_fig->setAttribute( 'class', 'wp-block-image' );
+			$img_fig->setAttribute( 'class', 'wp-block-image size-' . $size . ( $align ? ' align' . $align : '' ) );
 			
 			// Surround <figure> with block comments
 			$fragment = $dom->createDocumentFragment();
-			$fragment->appendChild( $dom->createComment( ' wp:image {"id":' . $img_id . ',"sizeSlug":"large"} ' ) );
+			$fragment->appendChild( $dom->createComment( ' wp:image ' . $json . ' ' ) );
 			$fragment->appendChild( $img_fig );
 			$fragment->appendChild( $dom->createComment( ' /wp:image ' ) );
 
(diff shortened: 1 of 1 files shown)


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-filter.php
```diff
@@ -522,17 +522,99 @@ public function filter_video_embeds( $html, $id, $attachment ) {
...
+	 * 
+	 * @return string
+	 */
+	protected function classic_image_to_block( $content ) {
+		if ( strpos( $content, '<img class=' ) === false ) {
+			return $content;
+		}
+
+		$dom = new \DOMDocument();                
+		@$dom->loadHTML( $content, LIBXML_HTML_NODEFDTD );
```
pereirinha, line 581: @dugajean 

I am wondering why we would want to silence the errors/warnings here.
Did you run into any while working on this?

dugajean: `loadHTML` complains when the HTML you try to load isn't a full-fledged HTML document (ie. no <html>, <body>, etc.). So a way to remedy this is by silencing it as the warnings aren't really saying anything important.

dugajean: More info here: https://stackoverflow.com/questions/9149180/domdocumentloadhtml-error

pereirinha: thanks for the explanation and comment in the code.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-filter.php
```diff
@@ -522,17 +522,102 @@ public function filter_video_embeds( $html, $id, $attachment ) {
...
+			$img_clone->setAttribute( 'class', 'wp-image-' . $img_id );
+
+			// Prep the <figure> node
+			$img_fig = $dom->createElement( 'figure' );
+			$img_fig->appendChild( $img_clone );
+			$img_fig->setAttribute( 'class', 'wp-block-image' );
+			
+			// Surround <figure> with block comments
+			$fragment = $dom->createDocumentFragment();
+			$fragment->appendChild( $dom->createComment( ' wp:image {"id":' . $img_id . ',"sizeSlug":"large"} ' ) );
```
pereirinha, line 608: @dugajean 

We should try to read the image size from the `$img` element — from the class name, eg. `size-thumbnail` —, and not assuming that we'll always use the `large` size.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-filter.php
```diff
@@ -522,17 +522,147 @@ public function filter_video_embeds( $html, $id, $attachment ) {
...
+		// Fetch image id
+		preg_match( '/wp-image-([0-9]+)/', $class_list, $matches );
+		$results[] = $id = isset( $matches[1] ) ? (int) $matches[1] : null;
+
+		// Fetch image size
+		$results[] = $size = 'large';
+		if ( false !== strpos( $class_list, 'size-' ) ) {
+			preg_match( '/size-([A-Za-z0-9-_]+)/', $class_list, $matches );
+
+			$all_sizes = get_intermediate_image_sizes();
```
pereirinha, line 643: @dugajean 

What you're trying to do here is great and it makes sense. However, this function on VIP will return you an [empty array](https://github.com/WordPress/WordPress-Coding-Standards/commit/e7707b2338e7823a2d14cfbdca8bdc88488d6081#diff-1034de57626216004cbb6c6e1707f313R244-R250).

I'd say to be ok to implement the same logic of the [function](https://developer.wordpress.org/reference/functions/get_intermediate_image_sizes/), as `wp_get_additional_image_sizes()` seems to be a safe method.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-filter.php
```diff
@@ -522,17 +522,147 @@ public function filter_video_embeds( $html, $id, $attachment ) {
...
+
+		// Fetch image alignment
+		$results[] = $align = null;
+		if ( false !== strpos( $class_list, 'align' ) && false === strpos( $class_list, 'alignnone' ) ) {
+			preg_match( '/align([a-z]+)/', $class_list, $matches );
+			$results[2] = $align = isset( $matches[1] ) ? $matches[1] : null;
+		}
+
+		// Prep json for block comment
+		$results[] = json_encode(
```
pereirinha, line 655: Can we use here `wp_json_encode()` instead?

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-filter.php
```diff
@@ -522,17 +522,147 @@ public function filter_video_embeds( $html, $id, $attachment ) {
...
+		$results[] = $align = null;
+		if ( false !== strpos( $class_list, 'align' ) && false === strpos( $class_list, 'alignnone' ) ) {
+			preg_match( '/align([a-z]+)/', $class_list, $matches );
+			$results[2] = $align = isset( $matches[1] ) ? $matches[1] : null;
+		}
+
+		// Prep json for block comment
+		$results[] = json_encode( 
+			array(
+				'id' 	   => $id,
```
pereirinha, line 657: There are here missed spaces and tabs in this indentation.

PR #133, opened 2020-07-23
Feature/unique-filenames

DavidCramer:


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
```diff
@@ -172,13 +172,58 @@ public function get_signature( $post_id ) {
...
+		$cld_folder = trailingslashit( $settings['sync_media']['cloudinary_folder'] );
+		$file       = get_attached_file( $attachment_id );
+		$file_info  = pathinfo( $file );
+		$public_id  = $cld_folder . $prefix . $file_info['filename'];
+		// Get the type.
+		if ( wp_attachment_is( 'image', $attachment_id ) ) {
+			$type = 'image';
+		} elseif ( wp_attachment_is( 'video', $attachment_id ) ) {
+			$type = 'video';
+		} elseif ( wp_attachment_is( 'video', $attachment_id ) ) {
```
pereirinha, line 193: Don't we have here a typo? I think that the previous check was already for `video`.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
```diff
@@ -172,13 +172,58 @@ public function get_signature( $post_id ) {
...
+			$type = 'audio';
+		} else {
```
pereirinha, line 195: Wondering if we shouldn't think about adding support for other types. I'm thinking about raw or text. This would be to handle the SVGs and others that might be in the pipeline.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
```diff
@@ -172,13 +172,58 @@ public function get_signature( $post_id ) {
...
+	 * @param int $attachment_id The attachment ID for the new public ID.
+	 *
+	 * @return string|null
+	 */
+	public function generate_public_id( $attachment_id, $prefix = null ) {
+		$settings   = $this->plugin->config['settings'];
+		$cld_folder = trailingslashit( $settings['sync_media']['cloudinary_folder'] );
+		$file       = get_attached_file( $attachment_id );
+		$file_info  = pathinfo( $file );
+		$public_id  = $cld_folder . $prefix . $file_info['filename'];
```
pereirinha, line 187: I wonder if suffix wouldn't be a better approach to make the public id unique. The reason for that has to do with sorting.
What do you think?

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
DavidCramer: Yes, but at this stage, WordPress doesn't support it, so week need to stick to what we have for now.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
DavidCramer: @pereirinha On the SEO level, I think that having the name at the end will better benefit the SEO rating.

pereirinha: @DavidCramer 

That's interesting. Do you have an article that you can refer to? This way, we document this decision properly, as for us humans, there are other preferred ways to see filenames.

DavidCramer: @pereirinha - Well, it was the post on seo file names in Cloudinary's docs. It just had descriptive names as the last item.
However, it's minor and the sorting is bigger issue, so let's move it to a suffix. It's cleaner. And even then, we can alter this with the public_id filter anyway.

DavidCramer:
@pereirinha - Try this one. However, we'll need to document the filter for the `public_id` better.

DavidCramer:
@pereirinha - Please give this another review. I've made some changes that make this way better and allows the public_id to be filtered freely.

PR #137, opened 2020-07-28
Cached fonts

pereirinha:
The previous cache strategy was useless as the file-loader is removing the query string on the build process.
It means that on sites that rely on CDNs to serve these static assets, the new icons in the coming release will not be displayed until the cache is flushed.
This new approach changes this and adds a unique piece to the file name. It also plays nicely with CDNs that ignores the query string as a cache buster.


pereirinha:
Hi Asi:

I understand that out of context, this might sound bad, which isn't the case.

This problem was identified when we added new icons. These icons are part of a custom font that we created for Cloudinary.

CDN services like Cloudflare or Cloudfront do cache static assets like fonts and the strategy we had before to bust the cache was dependent on a query string, like cloudinary.ttf?hdjeiri. Changing this string should force refreshing the cached asset.

However, some CDNs are ignoring this so they'll still serve the cached version, even if the the string is changed. On top of this, webpack was stripping out the query string in the compilation time, so we wouldn't even have that.

The reason why this wasn't a problem before, is because this is the first time we are adding new icons since the release.

Let me know if you've more questions about this.

Thanks. 

Marco

PR #145, opened 2020-07-30
Feature/revise sync meta

DavidCramer:


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-connect.php
```diff
@@ -256,6 +256,15 @@ public function get_credentials() {
 		return $this->credentials;
 	}
 
+	/**
+	 * Get the cloud name if set.
+	 *
+	 * @return bool|mixed
+	 */
+	public function get_cloud_name() {
+		return $this->credentials['cloud_name'] ? $this->credentials['cloud_name'] : false;
```
dugajean, line 265: It would be cool if we returned `null` instead of `false` so we can avoid mixing the types here

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
```diff
@@ -634,33 +675,47 @@ public function get_public_id( $attachment_id ) {
...
+	/**
+	 * Get a Cloudinary ID which includes the file format extension.
 	 *
 	 * @param int $attachment_id The Attachment ID.
 	 *
 	 * @return string A cloudinary id.
 	 */
 	public function get_cloudinary_id( $attachment_id ) {
 
+		$public_id = false;
```
dugajean, line 704: Like above, I'd prefer if `null` was used here instead. What do you think?

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
```diff
@@ -634,33 +675,47 @@ public function get_public_id( $attachment_id ) {
...
+	public function has_public_id( $attachment_id ) {
+		return ! empty( $this->get_post_meta( $attachment_id, Sync::META_KEYS['public_id'], true ) );
+	}
+
+	/**
+	 * Get a Cloudinary ID which includes the file format extension.
 	 *
 	 * @param int $attachment_id The Attachment ID.
 	 *
 	 * @return string A cloudinary id.
```
dugajean, line 700: Add the possible `|false` (or `|null` if we change it) to the docblock.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
```diff
@@ -1027,7 +1081,7 @@ public function down_sync_asset() {
 			}
 			$transformations = $this->get_transformations_from_string( $url );
 			if ( ! empty( $transformations ) ) {
-				$sync_key                .= wp_json_encode( $transformations );
+				$sync_key                 .= wp_json_encode( $transformations );
```
dugajean, line 1084: I think we can remove this space 😉

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
```diff
@@ -1312,15 +1369,130 @@ public function delete_post_meta( $post_id, $key ) {
...
+	 * @param int $attachment_id The attachment ID.
+	 *
+	 * @return array
+	 */
+	public function get_breakpoint_options( $attachment_id ) {
+		// Add breakpoints if we have an image.
+		$breakpoints     = array();
+		$has_breakpoints = $this->plugin->config['settings']['global_transformations']['enable_breakpoints'];
+
+		if ( 'off' !== $has_breakpoints && wp_attachment_is_image( $attachment_id ) ) {
```
dugajean, line 1384: You can turn this line into:
```suggestion
		if ( $has_breakpoints && 'off' !== $has_breakpoints && wp_attachment_is_image( $attachment_id ) ) {
```
And then remove the `if` on line 1395.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
```diff
@@ -1312,15 +1369,130 @@ public function delete_post_meta( $post_id, $key ) {
...
+			// Get meta image size if non exists.
+			if ( empty( $meta ) ) {
+				$meta          = array();
+				$imagesize     = getimagesize( get_attached_file( $attachment_id ) );
+				$meta['width'] = $imagesize[0] ? $imagesize[0] : 0;
+			}
+			$max_width = $this->get_max_width();
+			$settings  = $this->plugin->config['settings'];
+			// Add breakpoints request options.
+			if ( ! empty( $settings['global_transformations']['enable_breakpoints'] ) ) {
```
dugajean, line 1395: This check here might be redundant. See my comment above

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
```diff
@@ -119,66 +163,654 @@ public function is_synced( $post_id ) {
...
 	 */
-	public function generate_signature( $post_id ) {
-		$upload = $this->managers['push']->prepare_upload( $post_id );
-		// Check if has an error (ususally due to file quotas).
-		if ( is_wp_error( $upload ) ) {
-			$this->plugin->components['media']->get_post_meta( $post_id, self::META_KEYS['sync_error'], $upload->get_error_message() );
+	public function generate_signature( $attachment_id, $cache = true ) {
+		static $signatures = array(); // cache signatures.
+		if ( ! empty( $signatures[ $attachment_id ] ) && true === $cache ) {
+			$return = $signatures[ $attachment_id ];
```
dugajean, line 174: We can directly return here: `return $signatures[ $attachment_id ]`

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
```diff
@@ -119,66 +163,654 @@ public function is_synced( $post_id ) {
...
+		} else {
+			$return = $this->sync_base( $attachment_id );
+			if ( ! is_wp_error( $return ) ) {
+				$return                       = array_map(
```
dugajean, line 178: The spaces here might not be required

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
```diff
@@ -119,66 +163,654 @@ public function is_synced( $post_id ) {
...
-	public function get_signature( $post_id ) {
+	public function get_signature( $attachment_id, $cached = true ) {
 		static $signatures = array(); // Cache signatures already fetched.
 
-		$return = false;
-		if ( ! empty( $signatures[ $post_id ] ) ) {
-			$return = $signatures[ $post_id ];
+		$return = array();
+		if ( ! empty( $signatures[ $attachment_id ] ) && true === $cached ) {
+			$return = $signatures[ $attachment_id ];
```
dugajean, line 250: Same as above, we can return here and avoid the `else` nesting, making the code more readable.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
```diff
@@ -119,66 +163,654 @@ public function is_synced( $post_id ) {
...
+	 *
+	 * @return string The public ID.
+	 */
+	public function add_suffix_maybe( $public_id, $attachment_id, $suffix = null ) {
+
+		// Test if asset exists by calling just the head on the asset url, to prevent API rate limits.
+		$url         = $this->plugin->components['connect']->api->cloudinary_url( $public_id . $suffix );
+		$req         = wp_remote_head( $url, array( 'body' => array( 'rdm' => wp_rand( 100, 999 ) ) ) );
+		$asset_error = strtolower( wp_remote_retrieve_header( $req, 'x-cld-error' ) );
+		$code        = wp_remote_retrieve_response_code( $req );
```
dugajean, line 298: Very cool approach here!

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
```diff
@@ -119,66 +163,654 @@ public function is_synced( $post_id ) {
...
+	}
+
+	/**
+	 * Maybe add a suffix to the public ID if it's not unique.
+	 *
+	 * @param string      $public_id     The public ID to maybe add a suffix.
+	 * @param int         $attachment_id The attachment ID.
+	 * @param string|null $suffix        The suffix to maybe add.
+	 *
+	 * @return string The public ID.
```
dugajean, line 290:
```suggestion
	 * @return string|null The public ID.
```

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-sync-queue.php
```diff
@@ -63,7 +63,7 @@ public function __construct( \Cloudinary\Plugin $plugin ) {
 
 		$this->cron_frequency    = apply_filters( 'cloudinary_cron_frequency', 10 * MINUTE_IN_SECONDS );
 		$this->cron_start_offset = apply_filters( 'cloudinary_cron_start_offset', MINUTE_IN_SECONDS );
-
+		$this->threads           = apply_filters( 'cloudinary_queue_threads', array( 'thread_0', 'thread_1', 'thread_2' ) );
```
pereirinha, line 66: Can we have here docblock for these filters?

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-sync-queue.php
```diff
@@ -240,48 +262,122 @@ public function build_queue() {
...
-	 * @return void
+	 * @return array
 	 */
 	public function start_queue() {
-		$queue            = $this->get_queue();
-		$queue['started'] = $queue['last_update'] = current_time( 'timestamp' );
+		$queue = $this->get_queue();
+		if ( ! empty( $queue['processing'] ) ) {
+			// In case it stopped mid process, push back to the  first thread.
+			$queue['thread_0'] = array_merge( $queue['thread_0'], $queue['processing'] );
```
pereirinha, line 322: Shouldn't we do this is all threads?

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
```diff
@@ -119,66 +163,654 @@ public function is_synced( $post_id ) {
...
+
+				// Exists, check to see if this asset originally belongs to this ID.
+				if ( ! empty( $cld_asset['context'] ) && ! empty( $cld_asset['context']['custom'] ) && ! empty( $cld_asset['context']['custom']['wp_id'] ) ) {
+					$context_id = (int) $cld_asset['context']['custom']['wp_id'];
+				}
+
+				// Generate new ID only if context ID is not related.
+				if ( $context_id !== $attachment_id ) {
+					// Generate a new ID with a uniqueID prefix.
+					$suffix = '-' . uniqid();
```
dugajean, line 326: `uniqid` will not guarantee uniqueness and will fail because of the system's clock:
> This function does not guarantee uniqueness of return value. Since most systems adjust system clock by NTP or like, system time is changed constantly. Therefore, it is possible that this function does not return unique ID for the process/thread.

Providing the `more_entropy` parameter might help, but still doesn't guarantee uniqueness. I suggest we go for a stronger random string here, such as uuid v4 or something else - to ensure we're never generating the same string twice.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
```diff
@@ -119,66 +163,654 @@ public function is_synced( $post_id ) {
...
+		// Apply a default to ensure parts exist.s
+		$structs = array_map(
+			function ( $component ) {
+				$sync_default   = array(
+					'priority' => 50,
+					'callback' => '__return_null',
+				);
+				$status_default = array(
+					'state' => 'sync',
+					'note'  => __( 'Syncronizing asset with Cloudinary', 'cloudinary' ),
```
dugajean, line 475:
```suggestion
					'note'  => __( 'Synchronizing asset with Cloudinary', 'cloudinary' ),
```

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
```diff
@@ -119,66 +163,654 @@ public function is_synced( $post_id ) {
...
+					return $this->add_suffix_maybe( $public_id, $attachment_id, $suffix );
+				}
+			}
+		}
+
+		return $suffix;
+	}
+
+	/**
+	 * Get the sync_setup_base of part callbacks.
```
dugajean, line 338: I don't understand the purpose of this method based on this docblock. It would be awesome if you could elaborate this further 😄

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-upload-sync.php
```diff
@@ -111,15 +117,15 @@ function add_inline_action( $actions, $post ) {
 					'<a href="%s" aria-label="%s">%s</a>',
 					$action_url,
 					/* translators: %s: Attachment title. */
```
pereirinha, line 119: @DavidCramer 
We can also remove this comment.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-upload-sync.php
```diff
@@ -111,15 +117,15 @@ function add_inline_action( $actions, $post ) {
...
-					esc_attr( sprintf( __( 'Push to Cloudinary &#8220;%s&#8221;' ), 'asd' ) ),
+					esc_attr( __( 'Push to Cloudinary' ) ),
```
pereirinha, line 120: Couldn't we use `esc_attr__()` instead? Can we add here the plugin namespace for the translation?

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
```diff
@@ -119,66 +163,654 @@ public function is_synced( $post_id ) {
...
+
+	/**
+	 * Get the method to do a sync for the specified type.
+	 *
+	 * @param $type
+	 *
+	 * @return bool|mixed
+	 */
+	public function get_sync_method( $type ) {
+		$method = false;
```
dugajean, line 523: Can we have this be `null`?

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-upload-sync.php
```diff
@@ -111,15 +117,15 @@ function add_inline_action( $actions, $post ) {
...
 					__( 'Push to Cloudinary', 'cloudinary' )
```
pereirinha, line 121: Can we add escaping here as well?

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-upload-sync.php
```diff
@@ -111,15 +117,15 @@ function add_inline_action( $actions, $post ) {
...
 				);
 			} else {
 				$actions['cloudinary-push'] = sprintf(
 					'<a href="%s" aria-label="%s">%s</a>',
 					$action_url,
 					/* translators: %s: Attachment title. */
```
pereirinha, line 127: Same comments as above.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-upload-sync.php
```diff
@@ -142,14 +148,8 @@ public function handle_bulk_actions( $location, $action, $post_ids ) {
 		switch ( $action ) {
 			case 'cloudinary-push' :
 				foreach ( $post_ids as $post_id ) {
-					delete_post_meta( $post_id, Sync::META_KEYS['sync_error'] );
```
pereirinha, line 145: I am wondering if we shouldn't keep these for a while. I suspect that there will be a few `sync_error` keys for a while as we missed some cleanups previously. Perhaps this is not even the best place to proceed with this cleanup. Happy to get your ideas.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
```diff
@@ -122,16 +130,35 @@ public function __construct( Plugin $plugin ) {
...
+	/**
+	 * Get a resource type based on file.(Cloudinary v1 remove mime type in post data).
+	 *
+	 * @param \WP_Post|int $attachment_id The attachment ID or object.
+	 *
+	 * @return string
+	 */
+	public function get_media_type( $attachment_id ) {
+		$file = pathinfo( get_attached_file( $attachment_id ), PATHINFO_BASENAME );
+		$mime = wp_check_filetype( $file );
```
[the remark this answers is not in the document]
DavidCramer: The second argument for pathinfo specifies I'm only wanting the basename, so it's a string, not an array.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
```diff
@@ -634,33 +675,47 @@ public function get_public_id( $attachment_id ) {
...
+		$public_id = false;
 		// A cloudinary_id is a public_id with a file extension.
-		$public_id     = $this->get_public_id( $attachment_id );
-		$file          = get_attached_file( $attachment_id );
-		$info          = pathinfo( $file );
-		$cloudinary_id = $public_id . '.' . $info['extension'];
+		if ( $this->has_public_id( $attachment_id ) ) {
+			$public_id   = $this->get_public_id( $attachment_id );
+			$suffix_data = $this->get_post_meta( $attachment_id, Sync::META_KEYS['suffix'], true );
+			if ( is_array( $suffix_data ) && ! empty( $suffix_data['suffix'] ) && $suffix_data['public_id'] === $public_id ) {
```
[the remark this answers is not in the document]
DavidCramer: This is largely ignored, and being addressed with the new work.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
```diff
@@ -102,15 +145,16 @@ public function is_active() {
...
-		if ( ! empty( $signature ) && ! empty( $expecting ) && $expecting === $signature ) {
-			return true;
-		}
-
-		if ( $this->plugin->components['settings']->is_auto_sync_enabled() && apply_filters( 'cloudinary_flag_sync', '__return_false' ) && ! get_post_meta( $post_id, Sync::META_KEYS['downloading'], true ) ) {
-			update_post_meta( $post_id, Sync::META_KEYS['syncing'], true );
+		if ( ! is_wp_error( $expecting ) ) {
+			$signature = $this->get_signature( $post_id );
+			// Sort to align orders for comparison.
+			ksort( $signature );
```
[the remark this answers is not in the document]
DavidCramer: Yes, but this really helps with debugging 😂

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
```diff
@@ -119,66 +163,654 @@ public function is_synced( $post_id ) {
...
+	/**
+	 * Initialize the background sync on requested images needing to be synced.
+	 */
+	public function init_background_upload() {
+		if ( ! empty( $this->to_sync ) ) {
+
+			$threads    = $this->managers['push']->queue->threads;
+			$chunk_size = ceil( count( $this->to_sync ) / count( $threads ) ); // Max of 3 threads to prevent server overload.
+			$chunks     = array_chunk( $this->to_sync, $chunk_size );
+			$token      = uniqid();
```
dugajean, line 768: As said before, we shouldn't rely on this for true uniqueness.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
```diff
@@ -119,66 +163,654 @@ public function is_synced( $post_id ) {
...
+			),
+			'folder'      => array(
+				'generate' => array( $this->managers['media'], 'get_cloudinary_folder' ),
+				'sync'     => array(
+					'priority' => 10,
+					'callback' => array( $this->managers['upload'], 'upload_asset' ),
+				),
+				'status'   => array(
+					'state' => 'info syncing',
+					'note'  => function () {
```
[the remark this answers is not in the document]
DavidCramer: Yes. It serves more as an example of the note being able to be run of extended.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
```diff
@@ -119,66 +163,654 @@ public function is_synced( $post_id ) {
...
+	 * Add media state to display syncing info.
+	 *
+	 * @param array    $media_states List of the states.
+	 * @param \WP_Post $post         The current attachment post.
+	 *
+	 * @return array
+	 */
+	public function filter_media_states( $media_states, $post ) {
+
+		$status = apply_filters( 'cloudinary_media_status', array(), $post->ID );
```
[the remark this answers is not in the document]
DavidCramer: Yup. I've been trying to be consistent in adding docblocks to filters.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-sync-queue.php
DavidCramer: Yes, but it got cumbersome to have to split it up when there's generally only 1 or 2 that get stuck.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
DavidCramer: Agreed, this is largely untouched from the revert last week.
This is what I'm working on today and tomorrow.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-sync-queue.php
```diff
@@ -157,12 +179,9 @@ public function mark( $id, $type = 'done' ) {
...
 	 */
 	public function is_running() {
-		if ( false === $this->running ) {
-			$queue         = $this->get_queue();
-			$this->running = empty( $queue['started'] ) ? false : true;
-		}
+		$queue = $this->get_queue();
 
-		return $this->running;
+		return empty( $queue['started'] ) ? false : true;
```
dugajean, line 184:
```suggestion
		return ! empty( $queue['started'] );
```

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
```diff
@@ -1312,15 +1369,130 @@ public function delete_post_meta( $post_id, $key ) {
...
+		// Add breakpoints if we have an image.
+		$breakpoints     = array();
+		$has_breakpoints = $this->plugin->config['settings']['global_transformations']['enable_breakpoints'];
+
+		if ( 'off' !== $has_breakpoints && wp_attachment_is_image( $attachment_id ) ) {
+			$meta = wp_get_attachment_metadata( $attachment_id );
+			// Get meta image size if non exists.
+			if ( empty( $meta ) ) {
+				$meta          = array();
+				$imagesize     = getimagesize( get_attached_file( $attachment_id ) );
```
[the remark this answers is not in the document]
DavidCramer: @pereirinha  True, but at this point we are not wanting the size to be changed. We're simply wanting a conversion of what's already stored. But, I'll take a deeper look.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
DavidCramer: Actually, this was partly pulled from the old code and I didn't check it fully. line 1395 is already not needed since it shouldn't even get here if it is off. I'll revise this. Thanks for the catch!

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
DavidCramer: As I mentioned, I prefer having as few return points as possible to aid in code coverage and testing. It also helps with debugging since you know where something will return, you can track that return value to find where it changes.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
DavidCramer: IT's got spaces because it's directly inline with the line below it.
I'll add a comment explaining it's adding to the cache, which will remove the spaces.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
DavidCramer: Same as above. trying to maintain a single exit point to aid in code coverage.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
DavidCramer: Thanks :) - Although this will change a little as it was mostly ignored in this round, as I'm working on this part this week.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
DavidCramer: I would further that with `Synchronising` 😂

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
DavidCramer: This will change with the tickets I'm working on this week

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
```diff
@@ -122,16 +130,35 @@ public function __construct( Plugin $plugin ) {
...
+			$is_media    = in_array( $type, $media_types );
 		}
 
 		return $is_media;
 	}
 
+	/**
+	 * Get a resource type based on file.(Cloudinary v1 remove mime type in post data).
+	 *
+	 * @param \WP_Post|int $attachment_id The attachment ID or object.
```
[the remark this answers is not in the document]
DavidCramer: @pereirinha - Correct. I was expecting an object since `get_post_type` accepts both. But I've change it to be more consistent by having `attachment_id` for all methods. I'll change this to int.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-upload-sync.php
DavidCramer: @pereirinha - The `sync_error` is still stored in the internal meta. This was removing it from standard `post_meta`
It's still kept in `Media::update_post_meta`

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
DavidCramer: @dugajean - This is not really for uniqueness, but simply to have a key that references to a transient. Which only last 120 seconds. It also gets the thread ID attached to keep them separated.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-sync-queue.php
DavidCramer: @dugajean 🤦 😂

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
pereirinha: @dugajean @DavidCramer 

I'm not sure if we aren't overcomplicating. What will be the problem if the combination of the filename and suffix already exists in Cloudinary? The system will find the next available combination.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-sync-queue.php
pereirinha: @DavidCramer 

Not sure if I follow. Aren't these threads running in parallel and independent of each other? Perhaps we can discuss this is our next tech sync?

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-upload-sync.php
pereirinha: I also had in mind the syncing and downloading keys, which are stored as plain post metas

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
DavidCramer: @pereirinha - yes. But I'm reworking this so this isn't even going to be in. The method is unused and never called.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-sync-queue.php
dugajean: Did you guys figure this out in the sync?

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-upload-sync.php
```diff
@@ -327,12 +327,20 @@ public function add_suffix_maybe( $attachment_id, $suffix = null ) {
...
 				if ( ! empty( $cld_asset['context'] ) && ! empty( $cld_asset['context']['custom'] ) && ! empty( $cld_asset['context']['custom']['guid'] ) ) {
 					$context_guid = $cld_asset['context']['custom']['guid'];
 				}
 
+				// Check to see if this asset originally belongs to this ID using the older wp_id.
+				if ( ! empty( $cld_asset['context'] ) && ! empty( $cld_asset['context']['custom'] ) && ! empty( $cld_asset['context']['custom']['wp_id'] ) ) {
+					if( (int) $cld_asset['context']['custom']['wp_id'] === $attachment_id ) {
+						$context_guid = $attachment_guid;
+					}
+				}
```
dugajean, line 344:
```suggestion
				if ( 
					! empty( $cld_asset['context'] ) && 
					! empty( $cld_asset['context']['custom'] ) && 
					! empty( $cld_asset['context']['custom']['wp_id'] ) && 
					(int) $cld_asset['context']['custom']['wp_id'] === $attachment_id 
				) {
					$context_guid = $attachment_guid;
				}
```

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-upload-sync.php
```diff
@@ -329,27 +337,32 @@ public function add_suffix_maybe( $attachment_id, $suffix = null ) {
...
 				}
 
 				// Check to see if this asset originally belongs to this ID using the older wp_id.
-				if ( 
-					! empty( $cld_asset['context'] ) && 
-					! empty( $cld_asset['context']['custom'] ) && 
-					! empty( $cld_asset['context']['custom']['wp_id'] ) && 
-					(int) $cld_asset['context']['custom']['wp_id'] === $attachment_id 
+				if (
+					! empty( $cld_asset['context'] ) &&
```
pereirinha: @DavidCramer 

Since we are updating this piece of code, this check and the following are pointless, as the `! empty( $cld_asset['context']['custom']['wp_id'] )` takes care of all :)

DavidCramer: @pereirinha - This is no longer here- it's been removed already.

PR #154, opened 2020-08-10
Fix revise sync meta

pereirinha:


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/connect/class-api.php
```diff
@@ -230,6 +230,7 @@ function ( $item ) use ( $transformation_index ) {
 	public function cloudinary_url( $public_id, $args = array(), $size = array(), $clean = false ) {
 		$defaults = array(
 			'resource_type' => 'image',
+			'version'       => '',
```
DavidCramer, line 233: @pereirinha - can you change the default for the version to v1?
```suggestion
			'version'       => '',
```

PR #155, opened 2020-08-10
Feature/single source offload

DavidCramer:


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/js/src/components/sync.js
```diff
@@ -118,6 +118,16 @@ const Sync = {
 
 export default Sync;
 
+// Add it a trigger watch to stop deactivation.
+let triggers = document.getElementsByClassName( 'cld-deactivate' );
+[ ...triggers ].map( ( trigger ) => {
```
dugajean, line 123: Since you're not changing the contents of the array, better to use `forEach` here.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-download-sync.php
```diff
@@ -249,7 +249,6 @@ public function download_asset( $attachment_id, $source = null ) {
 			// Generate signatures.
 			$this->sync->set_signature_item( $attachment_id, 'options' );
 			$this->sync->set_signature_item( $attachment_id, 'cloud_name' );
-			$this->sync->set_signature_item( $attachment_id, 'suffix' );
```
dugajean, line 252: Why are we removing this line? Is it relevant to the feature?

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-storage.php
```diff
@@ -0,0 +1,242 @@
...
+		$this->plugin   = $plugin;
+		$this->sync     = $this->plugin->components['sync'];
+		$this->media    = $this->plugin->components['media'];
+		$this->download = $this->sync->managers['download'];
+
+		add_action( 'cloudinary_register_sync_types', array( $this, 'setup' ), 20 );
+	}
+
+	/**
+	 * Generate a signature for this sync type.s
```
dugajean, line 90:
```suggestion
	 * Generate a signature for this sync type.
```

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-storage.php
```diff
@@ -0,0 +1,242 @@
...
+	 *
+	 * @return string
+	 */
+	public function generate() {
```
dugajean, line 94: Maybe this should be renamed to `generate_signature` for clarity

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-storage.php
```diff
@@ -0,0 +1,242 @@
...
+				$url = $this->media->cloudinary_url( $attachment_id, 'full', array( array( 'effect' => 'blur:100', 'quality' => $this->settings['low_res'] . ':440' ) ), null, false, true );
+				break;
+			case 'dual_full':
+				$state = $this->media->get_post_meta( $attachment_id, Sync::META_KEYS['storage'], true );
+				if ( ! empty( $state ) && 'dual_full' !== $state ) {
+					// Only do this is it's changing a state.
+					$url = $this->media->cloudinary_url( $attachment_id, '', array(), null, false, false );
+				}
+				break;
+		}
```
dugajean, line 119: We need to pick a `default` state or we open the system up for breaking (ie. admin with malicious intent)

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-storage.php
```diff
@@ -0,0 +1,242 @@
...
+			$this->download->download_asset( $attachment_id, $url );
+		}
+
+		$this->sync->set_signature_item( $attachment_id, 'storage' );
+		$this->sync->set_signature_item( $attachment_id, 'breakpoints' );
+		$this->media->update_post_meta( $attachment_id, Sync::META_KEYS['storage'], $this->settings['offload'] ); // Save the state.
+	}
+
+	/**
+	 * @param $attachment_id
```
dugajean, line 133: Description missing here

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-storage.php
```diff
@@ -0,0 +1,242 @@
...
+			case 'cld':
+				$note = __( 'Removing local copies.', 'cloudinary' );
+				break;
+			case 'dual_low':
+				$note = __( 'Reducing local resolutions', 'cloudinary' );
+				break;
+			case 'dual_full':
+				$note = __( 'Rebuilding local copies.', 'cloudinary' );
+				break;
+		}
```
dugajean, line 161: What if we want to add a new option in the future? That'd mean that we have to find all these switch statements to add logic to. As a minimal solution, we could have an array as a class property with the offload options, something like this:

```
protected $offload_options = [
    'cld'       => [
      'handler' => function ( $attachment_id ) {
        $this->remove_local_assets( $attachment_id );
      },
      'note'    => __( 'Removing local copies.', 'cloudinary' )
    ],
    'dual_low   => [
      'handler' => function ( $attachment_id ) {
        return $this->media->cloudinary_url( $attachment_id, 'full', array( array( 'effect' => 'blur:100', 'quality' => $this->settings['low_res'] . ':440' ) ), null, false, true );
      },
      'note'    => __( 'Reducing local resolutions', 'cloudinary' );
    ],
    'dual_full' => [
      'handler' => function ( $attachment_id ) {
        $state = $this->media->get_post_meta( $attachment_id, Sync::META_KEYS['storage'], true );

	if ( ! empty( $state ) && 'dual_full' !== $state ) {
		// Only do this is it's changing a state.
		return $this->media->cloudinary_url( $attachment_id, '', array(), null, false, false );
	}

        return null;
      },
      'note'    => __( 'Rebuilding local copies.', 'cloudinary' )
    ]
];
```

Then you can simple use it as:

```
$url = $this->offload_options[ $this->settings['offload'] ]['handler']( $attachment_id ); // use call_user_func here instead
```

or 

```
$note = $this->offload_options[ $this->settings['offload'] ]['note'];
```

Like said, this is the minimal extensible method that I could think of. Ideally, each option would be its own class, but that's a whole topic on its own 😄.

What do you think?

DavidCramer: @dugajean - that's a great point... I initially wanted this to be an offloading hook were anyone can hook into and offload how they want.
But due to timeframe I had to hard code it.

I think for now, we should focus on the ticket, and totally link this up with extending it later. Specifically CLOUD-366

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-storage.php
DavidCramer: ARH! good catch!

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-download-sync.php
DavidCramer: @dugajean - Suffix was removed as a sync type and made inline with the upload.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/js/src/components/sync.js
DavidCramer: @dugajean - please make a suggestion on this. My JS knowledge is lacking and this is the best way I know how to do it :)

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-storage.php
DavidCramer: Not a bad suggestion.

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/sync/class-storage.php
DavidCramer: good catch 🤦

DavidCramer:
> Overall, very very good implementation! It's a shorter PR than I thought it would be 👍. I have a few architectural concerns for future extensibility, but it shouldn't be too hard to fix.
> 
> Once addressed, I'll give this PR quick a test locally.

Yup, the new sync system makes things a lot simpler. We can easily add new sync types ( which is why this is a new class, as to add an example). Overall, this is a good direction I think.

dugajean:
@DavidCramer I switched to "Cloudinary Only", but this time the images aren't getting deleted at all.

DavidCramer:
> @DavidCramer I switched to "Cloudinary Only", but this time the images aren't getting deleted at all. There are also no messages in the media page.

@dugajean - Check that you use new images. The last one had an issue with components not loading, but the metadata would have been set. So it's probably not doing something because it's stuck in the wrong state.
I'll check again as well, just to be sure.

pereirinha (approved):
@DavidCramer 

I'm happy with this.
Please check with @dugajean his last few comments and confirm that's a matter of bad data.

PR #159, opened 2020-08-11
add check that a synctype is still available

DavidCramer:
@pereirinha - the metadata on whats saved (pending, syncing etc) may be outdated. If there is no sync-type needed due to the signature being in sync and up to date, don't show meta data notices, since it's no longer relevant.

### DIFF
--- a/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
+++ b/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
@@ -610,7 +610,7 @@ class Sync implements Setup, Assets {
 	 */
 	public function filter_status( $status, $attachment_id ) {
 
-		if ( $this->been_synced( $attachment_id ) && $this->is_pending( $attachment_id ) && $this->get_sync_type( $attachment_id ) ) {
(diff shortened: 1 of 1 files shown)


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
```diff
@@ -610,7 +610,7 @@ public function validate_sync_type( $type, $attachment_id ) {
 	 */
 	public function filter_status( $status, $attachment_id ) {
 
-		if ( $this->been_synced( $attachment_id ) || $this->is_pending( $attachment_id ) ) {
+		if ( $this->been_synced( $attachment_id ) && $this->is_pending( $attachment_id ) && $this->get_sync_type( $attachment_id ) ) {
```
dugajean, line 613: Calling `get_sync_type` outside the `if` may be a better idea to avoid calling the same function multiple times

DavidCramer: @dugajean - a good observation, but at this point, the `get_sync_type` will be using the cached version, meaning it will be run once on the page load. So it should represent the state at time of load.

PR #160, opened 2020-08-11
Feature/update upload icon

pereirinha:


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
```diff
@@ -343,7 +343,7 @@ public function setup_sync_base_struct() {
 				'generate' => 'get_attached_file',
 				'priority' => 5.1,
 				'sync'     => array( $this->managers['upload'], 'upload_asset' ),
-				'state'    => 'info',
+				'state'    => 'uploading',
```
DavidCramer, line 346: @pereirinha - shouldn't this be `info uploading` so that's it's blue?

DavidCramer: oh, unless you intend on it being orange... 🤔

pereirinha: @DavidCramer they are supposed to be orange, as in the original functionality

PR #163, opened 2020-08-12
Extend featured images

dugajean:
Allow overwriting transformations for featured image. Works on both classic editor and block editor.


dugajean:
The changes look great @pereirinha !

DavidCramer:
@pereirinha @dugajean - I'm closing this in exchange for a new PR (https://github.com/cloudinary/cloudinary_wordpress/pull/167) I made to solve this.

While there was nothing seriously wrong with this PR, it addressed the issue in the wrong direction, and caused the removal of all transformations of the source image. It also didn't take into account of the location in which the attachment was, so if the attachment was elsewhere, if it was in the same page, and is the attachment ID, then it would remove them. 

The new PR is essentially reworked from this, and removed the unnecessary changes to the builds tools.

PR #169, opened 2020-08-21
Add missing srcset to mismatched images

dugajean:
As described, this ticket will ensure that images that had a `-1` (or `-n`) appended by WP in order to avoid duplicate file names - will still have the `srcset` attribute applied to them.

Test instructions:
- Create a new post and add an image via the Cloudinary tab
- Add another image block
- Select the same image from the Cloudinary tab, but this time add some transformations to this image
- Save the post and observe that all the images will have `srcset` set in the frontend


pereirinha (approved):
Thanks @dugajean 
This looks great.

PR #170, opened 2020-08-25
edited downsync source

DavidCramer:


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
```diff
@@ -378,7 +378,7 @@ public function get_id_from_url( $url ) {
 	 *
```
dugajean, line 378: Need new `@param` for `$all`

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/js/src/components/widget.js
```diff
@@ -94,6 +94,12 @@ if ( wp.media && window.CLDN ) {
 											wp.Uploader.queue.reset();
 										}
 									}
+									if ( typeof asset.resync !== 'undefined' ) {
+										asset.resync.forEach( function( update_asset ){
+											let update_attach = attachment.get( update_asset.id );
+                                            update_attach.set( update_asset );
```
dugajean, line 100: Indentation got a little messy here 😆

DavidCramer: 🤣 what happened there!?

I'll fix it

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
```diff
@@ -1235,65 +1231,110 @@ public function balance_crop( $size, $shift_size ) {
...
+						$this->sync->set_signature_item( $id, 'storage', '' );
+						if ( $id !== $asset['attachment_id'] ) {
+							$resync[] = wp_prepare_attachment_for_js( $id );
+						}
+					}
+					// Use the primary ID as the main return, and add the resynced assets to that.
+					$return           = wp_prepare_attachment_for_js( $asset['attachment_id'] );
+					$return['resync'] = $resync;
+				} else {
+					$return              = wp_prepare_attachment_for_js( $asset['attachment_id'] );
```
dugajean, line 1333: Since this line is being duplicated, we can add it on top, outside of the of the `if`

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-sync.php
```diff
@@ -234,7 +234,7 @@ public function can_sync( $attachment_id, $type = 'file' ) {
 	public function get_sync_version( $attachment_id ) {
 		$version = $this->managers['media']->get_post_meta( $attachment_id, self::META_KEYS['plugin_version'], true );
 
-		return $version;
+		return $version !== $this->plugin->version;
```
dugajean, line 237: This is always gonna return a bool now. We should update the docblock and update the method name to something like `check_sync_version`. The current name doesn't really apply anymore 😄

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
DavidCramer: No we can't. I don't think to do that, but meets data is being changed, so it has to be as late as possible.

PR #172, opened 2020-08-27
CNAME parsing

dugajean:


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-connect.php
```diff
@@ -172,11 +178,21 @@ public function verify_connection( $data ) {
 
 		if ( ! empty( $result['message'] ) ) {
 			add_settings_error( 'cloudinary_connect', $result['type'], $result['message'], 'error' );
-
 			return $current;
 		}
 
-		add_settings_error( 'cloudinary_connect', 'connection_success', __( 'Successfully connected to Cloudinary.', 'cloudinary' ), 'updated' );
+		// Check if the given URL has a cname and store it if present.
+		if ( preg_match( '/(?:@\w+)\/([A-Za-z-]+\.cloudinary.com)/', $data['cloudinary_url'], $cname ) ) {
```
dugajean, line 185: @DavidCramer I'm not sure whether CNAMEs will always end with `.cloudinary.com`. That's just an assumption I made. I'll correct this if this isn't the case.

DavidCramer: @dugajean - CNAME can be any domain.

DavidCramer (changes requested):
Everything looks and works fine.
Just that regex to allow any domain needs to be fixed.

PR #182, opened 2020-09-02
Disconnect plugin in case of a disabled account

dugajean:


DavidCramer:
@dugi - The reason here why we couldn't see the ping, was because the credentials are not set yet. This will always be 404 since it's an unauthorised call.
 I should have paid better attention to the cname refactor. It bypassed the credentials so it wasn't really working. Sorry for not catching that. 😞 
Since it's related to the cname issue, I've merged it together with the smaller issues from RC2.
Which the PR is here https://github.com/cloudinary/cloudinary_wordpress/pull/187

PR #194, opened 2020-09-07
use storage to determine if a download should happen

DavidCramer:


pereirinha (approved):
@DavidCramer 

LGTM.
Great stuff

PR #201, opened 2020-09-11
Fit instead of scale for crops

dugajean:


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
```diff
@@ -497,7 +497,7 @@ public function get_crop( $url, $attachment_id ) {
 						'wpsize' => $size_name,
 						'width'  => $size['width'],
 						'height' => $size['height'],
-						'crop'   => $cropped ? 'fill' : 'scale',
+						'crop'   => $cropped ? 'fill' : 'fit',
```
DavidCramer, line 500: @dugajean - it was specified that we use scale for this, so we can't change it. We had a whole ticket on this.
The issue is that the size for the square doesn't have the `cropped` flag.
What we should do is on line 491 don't set the `$crop` as false, but rather check if the dimensions are in ratio.
```
$cropped = ! wp_image_matches_ratio( $meta['width'], $meta['height'], $size['width'], $size['height'] );
```

DavidCramer (changes requested):

dugajean:
@DavidCramer Thanks! Addressed this.

pereirinha (approved):
@dugajean 

LGTM. I'm happy to see this fixed 👍

DavidCramer (approved):

PR #204, opened 2020-09-16
Hotfix 2.2.1

DavidCramer:


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-filter.php
```diff
@@ -716,7 +729,12 @@ public function setup_hooks() {
 		$filter = $this;
 		array_map(
```
dugajean, line 730: We can use this chance to get rid of the `array_map` for looping here and use a basic loop control.

dugajean (approved):
LGTM

PR #210, opened 2020-09-21
check for modal class

DavidCramer:


pereirinha (approved):
@DavidCramer 

Looks good to me ;)

PR #211, opened 2020-09-21
Replace alert with red notice in preview

dugajean:


pereirinha (approved):
@dugajean 

LGTM.
Good stuff.

PR #212, opened 2020-09-22
replace old URL in content for other instances of the local URL

DavidCramer:


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-filter.php
```diff
@@ -378,6 +378,8 @@ public function filter_out_local( $content ) {
 			}
 
 			$content = str_replace( $asset, $new_tag, $content );
```
dugajean, line 380: This could be simplified to

```suggestion
			$content = str_replace( array( $asset, $url ), array( $new_tag, $cloudinary_url ), $content );
```

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/media/class-filter.php
```diff
@@ -378,6 +378,8 @@ public function filter_out_local( $content ) {
...
+			// Additional URL change for backgrounds etc..
+			$content = str_replace( $url, $cloudinary_url, $content );
```
dugajean, line 382: If the change suggested above is used, this line can be removed 😄

dugajean (approved):
Left a minor potential improvement which is not mandatory. Good catch with this!

pereirinha (approved):
@DavidCramer 

This looks good to me. I agree with @dugajean suggestion, but not required.
I'm merging for the time being.

PR #219, opened 2020-09-25
Reset folder

pereirinha:
Allows pushing an asset update — either via `Re-sync to Cloudinary` or the bulk action `Push to Cloudinary` — that resets the location of the asset in Cloudinary account to the folder set in WordPress.

### DIFF
--- a/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
+++ b/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
@@ -633,7 +633,7 @@ class Media implements Setup {
 		if ( false !== $previous_url ) {
 			return substr( $url, $previous_url );
 		}
-		if ( ! doing_action( 'wp_insert_post_data' ) && false === $this->in_downsize ) {
+		if ( ! doing_action( 'wp_insert_post_data' ) && false === $this->in_downsize && ! apply_filters( 'cloudinary_doing_upload', '__return_false' ) ) {
 			if ( $this->cloudinary_id( $attachment_id ) ) {
 				$url = $this->cloudinary_url( $attachment_id );
 			}
--- a/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
+++ b/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
@@ -633,7 +633,19 @@ class Media implements Setup {
 		if ( false !== $previous_url ) {
 			return substr( $url, $previous_url );
 		}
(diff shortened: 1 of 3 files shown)


cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
```diff
@@ -633,7 +633,7 @@ public function attachment_url( $url, $attachment_id ) {
 		if ( false !== $previous_url ) {
 			return substr( $url, $previous_url );
 		}
-		if ( ! doing_action( 'wp_insert_post_data' ) && false === $this->in_downsize ) {
+		if ( ! doing_action( 'wp_insert_post_data' ) && false === $this->in_downsize && ! apply_filters( 'cloudinary_doing_upload', '__return_false' ) ) {
```
dugajean, line 636: This filter needs a docblock!

dugajean: Also this filter will return the string `__return_false` (which would result in `true`). What you're looking for here is:

```suggestion
		if ( ! doing_action( 'wp_insert_post_data' ) && false === $this->in_downsize && apply_filters( 'cloudinary_doing_upload', false ) ) {
```

dugajean (commented):
Great work! Works as expected. Left a few comments to check out before moving forward. 👍

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
pereirinha: @dugajean 

Great catch with the default value. 👍

cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-media.php
```diff
@@ -633,7 +633,19 @@ public function attachment_url( $url, $attachment_id ) {
...
+			&& false === $this->in_downsize
+			/**
+			 * Filter doing upload.
+			 * If so, return the default attachment URL.
+			 *
+			 * @param bool Default false.
+			 *
+			 * @return bool
+			 */
+			&& ! apply_filters( 'cloudinary_doing_upload', false )
```
dugajean, line 647: One small correction here:

```suggestion
			&& apply_filters( 'cloudinary_doing_upload', false )
```

Otherwise we default to true. Thanks for addressing this 😁

pereirinha: @dugajean 

The default should be true. We should almost always resolve to serve the Cloudinary. However, in the case of uploading, we should bail any change.

DavidCramer (approved):
This looks good to me. It's a nice addition.

PR #223, opened 2020-10-03
Gallery block widget

dugajean:


pereirinha (changes requested):
@dugajean 

I think this is on the right track, great job.
However, there are still some rough edges that need to be polished.

assets/css/logo.svg
```diff
@@ -1,4 +1,3 @@
-<?xml version="1.0"?>
```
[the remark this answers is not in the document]
dugajean: Reverted this

assets/css/components/_brand.scss
```diff
@@ -1,10 +1,9 @@
 @font-face {
 	font-family : 'cloudinary';
-	src         : url('fonts/cloudinary.eot');
-	src         : url('fonts/cloudinary.eot?x#iefix') format('embedded-opentype'),
-	url('fonts/cloudinary.ttf') format('truetype'),
-	url('fonts/cloudinary.woff') format('woff'),
-	url('fonts/cloudinary.svg?x#cloudinary') format('svg');
```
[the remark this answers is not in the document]
dugajean: Restored this.

Reason for these small changes in the css/svg is because I had some issues in the beginning when settings up the new webpack config, so I was trying different things and forgot to restore them. Great catches!

package.json
```diff
@@ -1,101 +1,64 @@
 {
```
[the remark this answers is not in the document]
dugajean: Can you elaborate on this?

ui-definitions/tabs/settings-gallery.php
```diff
@@ -0,0 +1,213 @@
...
+			'type'      => 'select',
+			'choices'   => array(
+				'top'    => __( 'Top', 'cloudinary' ),
+				'right'  => __( 'Right', 'cloudinary' ),
+				'left'   => __( 'Left', 'cloudinary' ),
+				'bottom' => __( 'Bottom', 'cloudinary' ),
+			),
+			'default'   => 'none',
+			'condition' => array(
+				'zoom_trigger' => array( 'none', 'neq' ),
```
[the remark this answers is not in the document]
dugajean: Not gonna focus much on settings page problems since it's likely that they wont stay around

php/media/class-gallery.php
```diff
@@ -0,0 +1,323 @@
...
+	 * Setup hooks for the gallery.
+	 */
+	public function setup_hooks() {
+		add_action( 'wp_enqueue_scripts', array( $this, 'enqueue_gallery_library' ) );
+		add_filter( 'cloudinary_api_rest_endpoints', array( $this, 'rest_endpoints' ) );
+		add_action( 'enqueue_block_editor_assets', array( $this, 'block_editor_scripts_styles' ) );
+
+		if ( ! is_admin() ) {
+			// phpcs:disable
+			add_filter( 'woocommerce_single_product_image_thumbnail_html', function () { return ''; } );
```
pereirinha, line 319: Can we use instead ['__return_empty_string'](https://developer.wordpress.org/reference/functions/__return_empty_string/)?

php/media/class-gallery.php
```diff
@@ -0,0 +1,323 @@
...
+	 * Fetches image public id and transformations.
+	 *
+	 * @param array|int[]|array[] $images An array of image IDs or a multi-dimensional array with url and id keys.
+	 *
+	 * @return array
+	 */
+	public function get_image_data( array $images ) {
+		$image_data = array();
+
+		foreach ( $images as $index => $image ) {
```
pereirinha, line 252: What if we filter out images that aren't synced with Cloudinary?

php/media/class-gallery.php
```diff
@@ -0,0 +1,323 @@
...
+		$config  = $this->get_config();
+
+		if ( $product ) {
+			$config['mediaAssets'] = $this->get_image_data( $product->get_gallery_image_ids() );
+		}
+
+		wp_localize_script(
+			self::GALLERY_LIBRARY_HANDLE,
+			'cloudinaryGallery',
+			array( 'config' => wp_json_encode( $config ) )
```
pereirinha, line 181: do we need to encode the `$config`?

pereirinha (changes requested):
@dugajean 

It follows a couple more suggested ideas/fixes. But let's wait for internal definition on the road map before us fixing this.

php/media/class-gallery.php
dugajean: Yes, otherwise using it as an array directly we lose the the value types and everything becomes a string. I need integers and booleans to remain in tact.

php/media/class-gallery.php
dugajean: Thanks!

php/class-settings-page.php
```diff
@@ -334,8 +335,8 @@ class="cld-field regular-<?php echo esc_attr( $type ); ?>"
 						<?php checked( $key, $value ); ?>
 							value="<?php echo esc_attr( $key ); ?>"
 					/>
-					<label for="<?php echo esc_attr( $field['label_for'] . '_' . $key ); ?>"><?php echo esc_html( $option ); ?></label>
-					<?php
+					<label for="<?php echo esc_attr( $field['label_for'] . '-' . $key ); ?>"><?php esc_html_e( $option ); ?></label>
```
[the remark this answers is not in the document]
dugajean: The linter throws an error if I do what you suggested. We just need to escape everything in the attribute.

php/media/class-gallery.php
DavidCramer: @dugajean - We need to use `wp_add_inline_script` when adding code.

php/media/class-gallery.php
```diff
@@ -0,0 +1,321 @@
...
+
+		wp_enqueue_script(
+			'cloudinary-gallery-block-js',
+			$this->media->plugin->dir_url . 'assets/dist/block-gallery.js',
+			array( 'wp-blocks', 'wp-editor', 'wp-element', self::GALLERY_LIBRARY_HANDLE ),
+			$this->media->plugin->version,
+			true
+		);
+
+		wp_localize_script(
```
DavidCramer, line 210: @dugajean - Here also, lets use `wp_add_inline_script`

php/media/class-gallery.php
```diff
@@ -0,0 +1,321 @@
...
+	public function get_image_data( array $images ) {
+		$image_data = array();
+
+		foreach ( $images as $index => $image ) {
+			$image_id  = is_int( $image ) ? $image : $image['id'];
+			$image_url = is_int( $image ) ? $this->media->cloudinary_url( $image_id ) : $image['url'];
+
+			if ( ! $this->media->sync->is_synced( $image_id ) ) {
+				continue;
+			}
```
DavidCramer, line 254: @dugajean - This should be the first check done.

php/media/class-gallery.php
dugajean: @DavidCramer I wanna stay away from writing actual JS in PHP. This current way works really well. 😄

php/media/class-gallery.php
dugajean: Good call 👍

php/media/class-gallery.php
DavidCramer: @dugajean - sure it works, but this is a hack. We need to follow best practises which is to use the correct functions for the task.

pereirinha (commented):
@dugajean 

I left a few more comments.
Can you please look into them?

Many thanks.

php/media/class-video.php
```diff
@@ -7,6 +7,8 @@
 
 namespace Cloudinary\Media;
 
+use Cloudinary\Media;
```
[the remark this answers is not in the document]
dugajean: Here I'm importing the class Media, but the namespace within the `media` folder is `Cloudinary\Media`, which isn't the same as importing the media class. Kinda confusing, I agree.

js/src/gallery-block/index.js
```diff
@@ -39,22 +39,46 @@ registerBlockType( 'cloudinary/gallery', {
...
+		},
+		displayProps_columns: {
+			type: 'number',
+			default: 1,
+		},
+		themeProps_primary: {
+			type: 'string',
+			default: primaryColor,
+		},
 		themeProps_onPrimary: {
```
pereirinha, line 54: @dugajean 

Let's remove this property as it doesn't seem to be used anywhere.

php/media/class-gallery.php
```diff
@@ -54,7 +53,6 @@ class Gallery {
 		'carousel_thumbnail_height'         => 64,
 		'carousel_button_shape'             => 'radius',
 		'carousel_thumbnail_selected_style' => 'gradient',
-		'custom_settings'                   => array(),
```
pereirinha: @dugajean 

Removing this key generates a notice:
`Undefined index: custom_settings in /var/www/html/wp-content/plugins/cloudinary_wordpress/php/media/class-gallery.php on line 107`

If we don't need it here, do we need it there?

js/src/gallery-block/edit.js
```diff
@@ -53,6 +52,13 @@ import { generateId, showNotice } from './utils';
 
 const dot = new Dot( '_' );
 
+const PLACEHOLDER_TEXT = Platform.select( {
+	web: __(
```
pereirinha, line 56: @dugajean 

Do we need here i18n namespace?

pereirinha (commented):
@dugajean

Can you please check the remaining comments?
They are pretty small ;)

PR #227, opened 2020-10-06
Fix/deploy tooling

pereirinha:


composer.json
```diff
@@ -0,0 +1,25 @@
...
+	"config": {
+		"platform": {
+			"php": "5.6.20"
+		}
+	},
+	"scripts": {
+		"lint": [
+			"phpcs"
+		],
+		"lint-fix": [
```
dugajean, line 21: Can we consider `fix` (without the lint-) or `format` here?

dugajean (approved):
Looks good!

DavidCramer (approved):

composer.json
pereirinha: @dugajean 

I updated the action name.

PR #229, opened 2020-10-12
Import using format set in transformation

DavidCramer:
If the format is set in a transformation, import to WordPress using that format.


dugajean (approved):

dugajean:
LGTM!

pereirinha (approved):
Looks good to me!

PR #239, opened 2020-10-20
Fix/gif mp4

pereirinha:


php/media/class-filter.php
```diff
@@ -371,6 +376,52 @@ public function filter_out_local( $content ) {
...
+	 *
+	 * @param string $cloudinary_url
+	 * @param int    $attachment_id
+	 *
+	 * @return string|null
+	 */
+	protected function check_fmp4_presence( $cloudinary_url, $attachment_id ) {
+		$return = null;
+
+		if ( Utils::ends_with( $cloudinary_url, '.gif' ) ) {
```
DavidCramer, line 410: @dugajean @pereirinha  There's no reason to search for end like this since we have the ID we can simply check the mime_type:
```suggestion
		if ( 'image/gif' === get_post_mime_type( $attachment_id ) ) {
```

php/media/class-filter.php
```diff
@@ -371,6 +376,52 @@ public function filter_out_local( $content ) {
...
+			$all_transformations = Utils::flatten( $this->media->get_transformations_from_string( $cloudinary_url ) );
+
+			if ( in_array( 'mp4', $all_transformations ) ) {
+				$return = sprintf(
+					'<video class="cld-fluid wp-video-%s" autoplay controls src="%s">',
+					$attachment_id,
+					$cloudinary_url
+				);
+			}
```
DavidCramer, line 419: @dugajean @pereirinha  We have methods for finding the index of a transformation:
```suggestion
			$all_transformations = $this->media->get_transformations_from_string( $cloudinary_url );
			$index = $this->media->get_transformation( $all_transformations, 'fetch_format' );
			if( false !== $index && 'mp4' === $all_transformations[ $index ][ 'fetch_format' ] ) {
				$return = sprintf(
					'<video class="cld-fluid wp-video-%s" autoplay controls src="%s">',
					$attachment_id,
					$cloudinary_url
				);
			}
```

However, if we do go with having this as a setting, we can change the check for the presence of the `f_mp4` to the check of the setting.

php/class-utils.php
```diff
@@ -0,0 +1,58 @@
+<?php
+/**
+ * Utilities for Cloudinary.
+ *
+ * @package Cloudinary
+ */
+
+namespace Cloudinary;
+
+class Utils {
```
DavidCramer, line 10: @dugajean @pereirinha - As mentioned below, we have methods that already are able to do this, so this Class is not needed.

DavidCramer (changes requested):
@dugajean - Cool work on the the tag conversion. However, the `Utils` are not needed since we have functionality, which I pointed out in my comments.

php/class-utils.php
dugajean: The class is alright imo, it should stay there as these utils can come in handy at any time (even if not with this PR).

DavidCramer: @dugajean I agree that a utils library is useful, but there's no point in introducing code in a PR that is not part of the PR.
When we start refactoring in other places that does warrant a utils ( to reduce duplication) then that is the time to introduce it.

DavidCramer (changes requested):
@dugajean - Looks good on the GIF and transformations swapout, Just need to have the `Utils` class removed as it unused code and offers no value to this PR.

dugajean (approved):

PR #240, opened 2020-10-21
fix infinate loop

DavidCramer:
@dugajean @pereirinha f7c088bf5f4f1474d9846019b8e392bdef2b4625 intruduced a infinate loop by moving variable assignment outside of a while loop.
The first has a loop kickout, but causes the sync to work singularly, severely slowing the sync down.
The second was a full infinite loop and would lock up.


pereirinha (approved):
Thanks for catching this @DavidCramer

PR #247, opened 2020-10-23
Return a class component from the editor.MediaUpload filter for compatibility with other plugins

johnwatkins0:
This addresses this issue opened on the AMP plugin repo: https://github.com/ampproject/amp-wp/issues/5534 **Note that this issue completely breaks the block editor when both plugins are active**, making the two plugins incompatible.

The AMP plugin uses the `editor.MediaUpload` filter expecting to filter a class component. This is because the original component from Gutenberg core is a class component: 

https://github.com/WordPress/gutenberg/blob/95188199c8b8045322d7f75a2666d47ea6504ad2/packages/media-utils/src/components/media-upload/index.js#L227

The Cloudinary plugin, however, uses the same filter to return a function component. When Cloudinary's callback runs first, the subsequent callback is no longer able to extend the component as expected, causing an error that breaks the editor.

For reference, here's where the AMP plugin extends the original component, expecting a class: https://github.com/ampproject/amp-wp/blob/6ceca505b3be3a75f460a641f9bfb208c7bf67e6/assets/src/block-editor/components/with-media-library-notice.js#L32


DavidCramer (approved):
Thanks @johnwatkins0

PR #262, opened 2020-11-11
(Video) Separate JS in PHP into separate JS file and load it.

dugajean:


pereirinha (approved):
@dugajean 

Looks good to me.

PR #270, opened 2020-12-01
Fix/enqueue video static

DavidCramer:
Fixes an issue caused in #262 - the register assets should use static params since the object might not be active.

Screenshot 2020-12-01 at 16 29 54

### DIFF
--- a/php/media/class-video.php
+++ b/php/media/class-video.php
@@ -395,7 +395,7 @@ class Video {
 		wp_register_style( 'cld-player', 'https://unpkg.com/cloudinary-video-player@' . self::PLAYER_VER . '/dist/cld-video-player.min.css', null, self::PLAYER_VER );
 		wp_register_script( 'cld-core', 'https://unpkg.com/cloudinary-core@' . self::CORE_VER . '/cloudinary-core-shrinkwrap.min.js', null, self::CORE_VER, true );
 		wp_register_script( 'cld-player', 'https://unpkg.com/cloudinary-video-player@' . self::PLAYER_VER . '/dist/cld-video-player.min.js', array( 'cld-core' ), self::PLAYER_VER, true );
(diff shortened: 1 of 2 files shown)


php/media/class-video.php
```diff
@@ -395,7 +395,7 @@ public function register_scripts_styles() {
 		wp_register_style( 'cld-player', 'https://unpkg.com/cloudinary-video-player@' . self::PLAYER_VER . '/dist/cld-video-player.min.css', null, self::PLAYER_VER );
 		wp_register_script( 'cld-core', 'https://unpkg.com/cloudinary-core@' . self::CORE_VER . '/cloudinary-core-shrinkwrap.min.js', null, self::CORE_VER, true );
 		wp_register_script( 'cld-player', 'https://unpkg.com/cloudinary-video-player@' . self::PLAYER_VER . '/dist/cld-video-player.min.js', array( 'cld-core' ), self::PLAYER_VER, true );
-		wp_enqueue_script( 'cld-video-init', $this->media->plugin->dir_url . 'js/video-init.js', array( 'cld-player' ), $this->media->plugin->version, true );
+		wp_enqueue_script( 'cld-video-init', CLDN_URL . 'js/video-init.js', array( 'cld-player' ), self::CORE_VER, true );
```
pereirinha, line 398: @DavidCramer 

Wondering if we wouldn't need to bring back the `static` declaration for `register_scripts_styles()`. Since it isn't right, we shouldn't use `self`, right?

pereirinha (commented):
@DavidCramer 

This looks great and it's a good catch.
Just left you a comment.

php/media/class-video.php
DavidCramer: Not needed. I removed the call that needed it as static, but it's a good example of available static values.b

pereirinha (approved):
Thanks for checking my concerns @DavidCramer.

Issue #281, opened 2020-12-14
1000 items per itteration seems like too many


pattonwebz:
In testing uploading a large library of images and some videos I have found that the process becomes unresponsive. The site has a significant amount of memory, CPU and bandwidth available.

I would like a filter to exist to reduce the number of items per itteration to a number I am more comfortable with on the initial production push.

https://github.com/cloudinary/cloudinary_wordpress/blob/31a7ce2510cd8fc3e6b459283648f40075a5fc27/php/sync/class-sync-queue.php#L255

Vdeub-cloudinary:
Hi @pattonwebz, I am closing this issue as it has been solved in the ticket 137946 you raised

PR #282, opened 2020-12-14
Wrap the `posts_per_page` paramiter of sync query filterable

pattonwebz:
Adds `cloudinary_sync_limit` filter around the value used for the `posts_per_page` part of the sync query so it can be reduced or increased depending on the needs of the site sync is being run on.

Fixes: #281

### DIFF
--- a/php/sync/class-sync-queue.php
+++ b/php/sync/class-sync-queue.php
@@ -252,7 +252,7 @@ class Sync_Queue {
 			'post_type'           => 'attachment',
 			'post_mime_type'      => array( 'image', 'video' ),
 			'post_status'         => 'inherit',
-			'posts_per_page'      => 1000, // phpcs:ignore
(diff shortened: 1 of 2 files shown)


pereirinha:
@pattonwebz 

I'm closing this one as the sync was improved in a different direction.
If you'd like to give it a spin around, please check out the latest `develop` branch.

You'll find 2 filters, `cloudinary_queue_threads` and `cloudinary_autosync_threads`, and they expect an integer in return — 2 and 1 are the default values for these. They define the number of concurrent workers to process the bulk sync and the images requested via auto-sync. They define how quickly you can resolve your queues, and you can tweak them pending your system's capacity.

Lastly, the bulk sync will no longer be limited to 1000 assets. Instead, you'll be able to sync all the pending assets in a single click.

Thank you for your contribution, and I hope we can support you better in your next PR.

PR #289, opened 2020-12-23
Add support for conditional global transformations

pereirinha:


DavidCramer (approved):
Looks good.
I like the additional method here.

PR #291, opened 2020-12-24
Fix/sorting issue

DavidCramer:


php/settings/class-setting.php
```diff
@@ -881,13 +878,44 @@ public function create_setting( $slug, $params = array(), $parent = null ) {
...
-		$this->sort_settings();
+
+		// Get the position in which to insert the new setting.
+		$index = $this->get_position_index( $setting->get_param( 'priority', 10 ) );
+
+		$new_setting = array(
+			$setting->get_slug() => $setting,
+		);
+
+		// Add the new setting at the index based on the priority position.
```
pereirinha, line 891: @DavidCramer 

I'm approving this PR as is, but I'd like us to add here a note explaining the issues we had with PHP 5.6 and `uasort`.

pereirinha (approved):
As mentioned, I'm approving and merging this PR. But I'd like us to document the sorting solution.

PR #295, opened 2021-01-04
Sync UI component

DavidCramer:
Upgrades the sync system to work in the new UI.


php/settings/class-setting.php
```diff
@@ -456,8 +456,23 @@ protected function register_setting() {
...
+			$current_value = $setting->get_value();
+			$new_value     = $setting->get_component()->sanitize_value( $value );
+			/**
+			 * Filter the value before saving a setting.
+			 *
+			 * @param mixed   $new_value     The new setting value.
+			 * @param mixed   $current_value The setting current value.
+			 * @param Setting $value         The setting object.
+			 */
+			$new_value = apply_filters( "settings_save_setting_{$slug}", $new_value, $current_value, $setting );
```
pereirinha, line 470: @DavidCramer 

I'm wondering if we should prepend these filters with `cloudinary` as if a namespace?

php/ui/class-component.php
```diff
@@ -762,19 +762,21 @@ public static function init( $setting ) {
...
 		// Final check if type is callable component.
 		if ( ! is_string( $type ) || ! self::is_component_type( $type ) ) {
-			// Set to a default HTML component if not found.
-			$type = 'html';
 			// Check what type this component needs to be.
 			if ( is_callable( $type ) ) {
 				$setting->set_param( 'callback', $type );
 				$setting->set_param( 'type', 'custom' );
 				$type = 'custom';
+			} else {
```
pereirinha, line 772: Not quite relevant, but IMHO, I prefer how it was. Setting a default that can be conditionally overridden. As much as possible, I try to avoid `else` conditions. As I mentioned, this is more of a personal preference and not a change request.

pereirinha (changes requested):
@DavidCramer 

I faced a fatal due to type declaritions.
I also added a couple of other comments for you to consider.

Thank you ;)

php/ui/class-component.php
DavidCramer: @pereirinha - there is a specific reason I did it this way, one which is not obvious. I can show you later in our sync.

php/settings/class-setting.php
```diff
@@ -862,17 +902,48 @@ public function create_setting( $slug, $params = array(), $parent = null ) {
...
 	}
 
+	/**
+	 * Set an error/notice for a setting.
+	 *
+	 * @param string $error_code    The error code/slug.
+	 * @param string $error_message The error text/message.
+	 * @param string $type          The error type.
+	 */
+	public function add_error_notice( string $error_code, string $error_message, string $type = 'error' ) {
```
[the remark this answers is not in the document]
DavidCramer: oops - that was meant to be taken out. fixing.

pereirinha (approved):
Looks great @DavidCramer 
Thank you.

I only fixed a conflict so I could merge. Please, have another set of eyes on it to ensure that nothing was lost.

PR #301, opened 2021-01-07
custom notifications

DavidCramer:
@pereirinha - Please merge this last, as it has open PR's in it.


js/src/components/notices.js
```diff
@@ -3,30 +3,43 @@ const Notices = {
...
 			[ ...notices ].forEach( ( notice ) => {
-				notice.addEventListener( 'click', ( ev ) => {
-					// WordPress has an onClick that cannot be unbound.
-					// So, we have the click on our Notice, and act on the
-					// button as a target.
-					if ( 'notice-dismiss' === ev.target.className ) {
-						self._dismiss( notice );
-					}
-				} );
+				const dismiss = notice.getElementsByClassName(
```
pereirinha, line 8: @DavidCramer 

we could had used here as well `.querySelector`, and that should return you only 1 or 0 items. It wouldn't require de use of `[0]` ahead.

pereirinha (approved):
@DavidCramer 

Only found something that I'd do differently, but def, not a blocker.
I'm approving this as is.

Thanks ;)

PR #302, opened 2021-01-08
Only show toggle if woocommerce is installed

dugajean:


pereirinha (approved):
Thanks @dugajean 

This looks great.

PR #310, opened 2021-01-12
expect SEO based urls in _wp_attached_file

DavidCramer:


pereirinha (approved):
Looks great.
Thanks @DavidCramer

PR #313, opened 2021-01-13
Global transformations bug

dugajean:
To test:

- Transformations used in images: `e_blackwhite:50`
- Transformations used in videos: `e_saturation:-200`

### DIFF
--- a/php/media/class-global-transformations.php
+++ b/php/media/class-global-transformations.php
@@ -160,30 +160,7 @@ class Global_Transformations {
 	 * @param string $type    The default transformations type.
 	 *
 	 * @return array
-	 private function get_term_transformations( $term_id, $type ) {
-	 $meta_data = array();
-	 foreach ( $this->taxonomy_fields[ $type ] as $setting ) {
-	 $slug     = $setting->get_slug();
-	 $meta_key = self::META_ORDER_KEY . '_' . $slug;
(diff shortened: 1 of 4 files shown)


php/media/class-global-transformations.php
```diff
@@ -162,18 +162,14 @@ public function save_taxonomy_custom_meta( $term_id ) {
...
-		foreach ( $this->taxonomy_fields[ $type ] as $setting ) {
-			$slug     = $setting->get_slug();
-			$meta_key = self::META_ORDER_KEY . '_' . $slug;
-			$value    = get_term_meta( $term_id, $meta_key, true );
-			$setting->set_value( $value );
-			$meta_data[ $slug ] = $value;
+		foreach ( $keys as $key ) {
+			$meta_key          = self::META_ORDER_KEY . '_' . $key;
+			$meta_data[ $key ] = get_term_meta( $term_id, $meta_key, true );
 		}
```
DavidCramer, line 170: @dugajean - Thanks for looking into this.
However, the issue is that on line 170, it's setting the value.
This is fine, however, it's setting this value on at a global state, meaning that the actual stored global is being set to whatever the term_meta returns. This is why the globals are not working; it's being removed before it can use the value.

The reason it was being set, is so the setting on the edit term screen, will have the value from the term_meta.
This should only happen when in the context of editing or creating a term.

I have addressed this oversight in the original branch that introduced it. See #314

PR #314, opened 2021-01-13
Fix taxonomy transformations

DavidCramer:
Corrects the setting of the `image_freeform` value outside of the terms context.


pereirinha (approved):
Looks great.

PR #315, opened 2021-01-13
get get_url defect

DavidCramer:
A defect in over-registering settings resulting in a `null` component when requested, which made the `get_url` method to return null. Hence the URL being the same as the current page.

The result is to check that the setting hasn't already been set first.
Additionally, removed the direct calling to the `cloudinary_url` setting by name as well, as the the direct calling of `connect` setting. This would be a technical debt if the settings pages got moved around.
Instead, we use the primary URL setting key as a constant and fetch the settings option parent. 
This will always return the setting page, where the `get_url` will point to wherever the setting is located.


pereirinha (approved):
Thanks.
Looks good.

PR #324, opened 2021-01-13
Change parser from text/xml to text/html for gallery

dugajean:


pereirinha (approved):
This works great @dugajean 
Thank you.

PR #328, opened 2021-01-15
Fix gallery block inconsistencies & Refactor

dugajean:


pereirinha (approved):
Looking good @dugajean.

PR #329, opened 2021-01-15
Use render_block to prep gallery for rendering

dugajean:


php/media/class-gallery.php
```diff
@@ -398,6 +396,39 @@ protected function register_settings() {
...
+		$attributes['cloudName']   = $this->media->plugin->components['connect']->get_cloud_name();
+		unset( $attributes['selectedImages'], $attributes['customSettings'] );
+
+		ob_start();
+		?>
+<script>
+	(function () {
+		const attributes = JSON.parse( '<?php echo wp_json_encode( $attributes ); ?>' );
+		attributes.container = '.' + attributes.container;
+		cloudinary.galleryWidget( attributes ).render();
```
kasparsd, line 424: This kind of direct function call requires that the JS API is already loaded and defined somewhere above in the HTML DOM. Could we use a queue system instead where all of the widget config gets pushed to an array which is then parsed and loaded as widgets in the footer after the external API is loaded?

php/media/class-gallery.php
```diff
@@ -236,15 +243,6 @@ public function block_editor_scripts_styles() {
 			$this->media->plugin->version,
 			true
 		);
-
-		wp_localize_script(
```
kasparsd, line 240: Is this no longer necessary?

js/src/components/gallery-init.js
```diff
@@ -1,22 +1,20 @@
...
-) {
-	cloudinary.galleryWidget( cloudinaryGalleryConfig ).render();
-}
+	window.onload = () => {
+		if (
+			document.querySelector( '.woocommerce-page' ) &&
+			CLD_GALLERY_CONFIG &&
+			CLD_GALLERY_CONFIG?.mediaAssets?.length
+		) {
+			galleryWidget( CLD_GALLERY_CONFIG ).render();
```
kasparsd, line 17: Is this passing the global gallery config to all WooCommerce product pages?

js/src/gallery-block/save.js
```diff
@@ -1,28 +1,5 @@
...
-
-	return (
-		<div
-			className={ attributes.container }
-			data-cloudinary-gallery-config={ configString }
-		></div>
-	);
-};
+const Save = ( { attributes } ) => (
+	<div className={ attributes.container }></div>
```
kasparsd, line 2: Question: Does the `container` attribute contain the block instance reference that we can use to extract the block-specific attribute overrides?

php/media/class-gallery.php
```diff
@@ -102,6 +101,13 @@ class Gallery {
 	 */
 	protected $config = array();
 
+	/**
+	 * Where the gallery script should be enqueued. false = header; true = footer.
+	 *
+	 * @var bool
+	 */
+	protected $gallery_in_footer = false;
```
kasparsd, line 109: Could we keep this in the footer if we use the queue system which stores just the widget instances and their configs on the current page?

php/media/class-gallery.php
```diff
@@ -153,18 +159,18 @@ public function enqueue_gallery_library() {
 			self::GALLERY_LIBRARY_URL,
 			array(),
 			$this->media->plugin->version,
-			true
+			$this->gallery_in_footer
 		);
 
 		$json_config = wp_json_encode( $this->get_config() );
-		wp_add_inline_script( self::GALLERY_LIBRARY_HANDLE, "var cloudinaryGalleryConfig = JSON.parse( '{$json_config}' );" );
+		wp_add_inline_script( self::GALLERY_LIBRARY_HANDLE, "var CLD_GALLERY_CONFIG = JSON.parse( '{$json_config}' );" );
```
kasparsd, line 166: `wp_add_inline_script()` puts this JS _after_ the parent JS by default per https://developer.wordpress.org/reference/functions/wp_add_inline_script/ -- is that correct considering we need the widget defaults before we can render the widgets?

php/media/class-gallery.php
```diff
@@ -220,6 +226,7 @@ static function ( $dependency ) {
 	 * Register blocked editor assets for the gallery.
 	 */
 	public function block_editor_scripts_styles() {
+		$this->gallery_in_footer = true;
```
kasparsd, line 229: Question: What is the intention of changing the external JS placement?

php/media/class-gallery.php
```diff
@@ -398,6 +396,39 @@ protected function register_settings() {
...
+
+		$attributes                = Utils::expand_dot_notation( $block['attrs'], '_' );
+		$attributes                = array_merge( self::$default_config, $attributes );
+		$attributes['mediaAssets'] = $attributes['selectedImages'];
+		$attributes['cloudName']   = $this->media->plugin->components['connect']->get_cloud_name();
+		unset( $attributes['selectedImages'], $attributes['customSettings'] );
+
+		ob_start();
+		?>
+<script>
```
kasparsd, line 420: The indentation here appears to be different from the rest of the method -- is that intentional?

php/media/class-gallery.php
```diff
@@ -398,6 +396,39 @@ protected function register_settings() {
...
+	(function () {
+		const attributes = JSON.parse( '<?php echo wp_json_encode( $attributes ); ?>' );
```
kasparsd, line 422: What is the purpose of the `JSON.parse()` here? Could we assign the output of `wp_json_encode()` directory?

Screenshot 2021-01-15 at 22 36 53

Also, could we use `var` here just for compatibility with older browsers?

php/media/class-woocommercegallery.php
```diff
@@ -42,7 +42,7 @@ public function enqueue_gallery_library() {
 
 		if ( $assets ) {
 			$json_assets = wp_json_encode( $assets );
-			wp_add_inline_script( Gallery::GALLERY_LIBRARY_HANDLE, "cloudinaryGalleryConfig.mediaAssets = JSON.parse( '{$json_assets}' );" );
+			wp_add_inline_script( Gallery::GALLERY_LIBRARY_HANDLE, "CLD_GALLERY_CONFIG.mediaAssets = JSON.parse( '{$json_assets}' );" );
```
kasparsd, line 45: Same question here -- is there a specific reason for using `JSON.parse()`?

kasparsd (commented):
@dugajean Please review my questions and suggestion inline. Thanks!

php/media/class-gallery.php
dugajean: Found a clean workaround for this!

php/media/class-gallery.php
dugajean: Found a clean workaround for this!

php/media/class-gallery.php
dugajean: Found a clean workaround for this!

js/src/gallery-block/save.js
dugajean: This is mostly needed so that the gallery widget lib has a particular selector to target.

js/src/components/gallery-init.js
dugajean: Yep, that's what it's doing!

php/media/class-gallery.php
dugajean: Opted for using WP's `apiFetch`, which handles this part out of the box.

php/media/class-gallery.php
dugajean: That would require hooking into a high part of the DOM (ie. `wp_head`) in order to initialize an empty variable. I found a workaround for this by simply listening for the `load` event.

js/src/components/notices.js
```diff
@@ -12,11 +12,7 @@ const Notices = {
 					dismiss[ 0 ].addEventListener( 'click', ( ev ) => {
 						notice.style.height = notice.offsetHeight + 'px';
 						ev.preventDefault();
-						// Give the event a slight delay to allow the height to
```
pereirinha, line 15: @dugajean, I'd get @DavidCramer input on this one.
He told me that this was needed to be able to get the height of the notifications before the animation.

js/src/components/notices.js
```diff
@@ -25,11 +21,10 @@ const Notices = {
 	_dismiss( notice ) {
 		const token = notice.dataset.dismiss;
 		const duration = parseInt( notice.dataset.duration );
+
 		notice.classList.add( 'dismissed' );
-		notice.style.height = '0px';
```
pereirinha, line 29: @dugajean 

Same here.

pereirinha (commented):
I still want to go through the PR tomorrow when I'm fresh, but I leave here a concern.

js/src/components/notices.js
DavidCramer: @pereirinha Yes, I put in the delay/timeout to aid in calculating heights to allow the animation to fire correctly.
With this change, I can confirm the animations don't work anymore. There's also no reason to edit the notice JS in this PR as it's unrelated to the gallery.

js/src/components/gallery-init.js
```diff
@@ -1,22 +1,20 @@
...
-const configElements = document.querySelectorAll(
-	'[data-cloudinary-gallery-config]'
-);
+( () => {
+	/**
+	 * A way to catch the galleryWidget function which comes from the `cloudinary` object.
+	 * The `cloudinary` object is later overwritten by the video player library by Cloudinary,
+	 * which effectively removes the gallery widget library from the page.
+	 */
+	const { galleryWidget } = cloudinary;
```
kasparsd, line 9: Is this file going through a transpiler to be compiled for all browsers? Object destructuring isn't supported in IE, for example https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#browser_compatibility

js/src/components/gallery-init.js
```diff
@@ -1,22 +1,20 @@
...
-		cloudinary.galleryWidget( options ).render();
-	} );
-} else if (
-	document.querySelector( '.woocommerce-page' ) &&
-	cloudinaryGalleryConfig &&
-	cloudinaryGalleryConfig?.mediaAssets?.length
-) {
-	cloudinary.galleryWidget( cloudinaryGalleryConfig ).render();
-}
+	window.onload = () => {
```
kasparsd, line 11: Should we use `window.addEventListener()` here to avoid overriding all other callbacks associated with the `onload` event here?

js/src/components/gallery-init.js
```diff
@@ -1,22 +1,20 @@
...
+		if (
+			document.querySelector( '.woocommerce-page' ) &&
+			CLD_GALLERY_CONFIG &&
+			CLD_GALLERY_CONFIG?.mediaAssets?.length
```
kasparsd, line 15: Is this kind of syntax supported by the browsers we're targeting? Or is this going through a transpiler which adds the legacy browser support?

js/src/gallery-block/save.js
kasparsd: Thanks for confirming! Is it unique for each block?

php/media/class-gallery.php
```diff
@@ -398,6 +391,39 @@ protected function register_settings() {
...
+		$attributes['cloudName']   = $this->media->plugin->components['connect']->get_cloud_name();
+		unset( $attributes['selectedImages'], $attributes['customSettings'] );
+
+		ob_start();
+		?>
+		<script>
+			window.addEventListener( 'load', function () {
+				var attributes = <?php echo wp_json_encode( $attributes ); ?>;
+				attributes.container = '.' + attributes.container;
+				cloudinary.galleryWidget( attributes ).render();
```
kasparsd, line 419: Could we add a check here that ensures that `cloudinary.galleryWidget` is defined and available just to prevent any errors in the browser console if the library isn't loaded for some reason?

php/media/class-gallery.php
```diff
@@ -237,14 +237,7 @@ public function block_editor_scripts_styles() {
...
 
-		wp_localize_script(
-			'cloudinary-gallery-block-js',
-			'cloudinaryGalleryApi',
-			array(
-				'endpoint' => rest_url( REST_API::BASE . '/image_data' ),
-				'nonce'    => wp_create_nonce( 'wp_rest' ),
-			)
-		);
+		wp_add_inline_script( 'cloudinary-gallery-block-js', 'var CLD_REST_ENDPOINT = "/' . REST_API::BASE . '";' );
```
kasparsd, line 240: Is it OK that this gets added _after_ the `cloudinary-gallery-block-js` is loaded? Shouldn't it be before the script which uses this data? https://developer.wordpress.org/reference/functions/wp_add_inline_script/

php/media/class-woocommercegallery.php
kasparsd: @dugajean You marked this as resolved but left in `JSON.parse()` -- is that intentional? Can you please confirm the usage of `JSON.parse()` here?

kasparsd (commented):
@dugajean Please see my comments inline! Thanks!

js/src/components/gallery-init.js
pereirinha: Yes @kasparsd, Babel should be doing the transpile of the code. Also, we dropped support for IE :)

js/src/components/gallery-init.js
pereirinha: Good point.
@dugajean what do you think?

js/src/components/gallery-init.js
pereirinha: Babel should handle this.

js/src/components/notices.js
pereirinha: @dugajean 
Let's rever this change, please.

pereirinha:
@dugajean 

We are loading on every page the scripts from `enqueue_gallery_library`. Can we improve this by checking if the block is used, or if it's a Woo product page, and that we are replacing the gallery?

js/src/gallery-block/save.js
dugajean: Yep, I generate this random class name in edit.js.

php/media/class-gallery.php
dugajean: This works because the variable will be defined by the time it's needed, but your suggestion adds proper precautions for when we'd need the variable at an earlier point. Thanks!

js/src/components/notices.js
dugajean: That animation was very clunky and bad to look at. That's the reason I removed it.

dugajean: Plus the solution to use many `setTimeout` invocations is not very clean. This can be achieved with CSS animations if really necessary.

php/media/class-woocommercegallery.php
dugajean: @kasparsd Thanks for checking this. I was under the impression that I caught all of them 😄

js/src/components/notices.js
pereirinha: The question here has more to do with the scope of the PR.
I don't think that one should address multiple subjects. If this were a separated PR, it would help the discussion and alternatives approach.

php/media/class-gallery.php
```diff
@@ -398,6 +391,47 @@ protected function register_settings() {
...
+
+		$attributes['mediaAssets'] = $attributes['selectedImages'];
+		$attributes['cloudName']   = $this->media->plugin->components['connect']->get_cloud_name();
+		unset( $attributes['selectedImages'], $attributes['customSettings'] );
+
+		ob_start();
+		?>
+		<script>
+			window.addEventListener( 'load', function () {
+				if ( cloudinary?.galleryWidget ) {
```
pereirinha, line 423: @dugajean 
Given that this is not going through Babel, and that [optional changing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) is only supported in recent versions of browsers, I think we should be a bit more explicit here.

dugajean: 🤦🏻

pereirinha (approved):
Thanks @dugajean 
Great stuff.

PR #337, opened 2021-01-18
Add the default transformation crop for the Gallery

pereirinha:


DavidCramer (approved):
Looks good.
We might need to review this later, if it causes unwanted results where the user wants full sizes.

pereirinha:
@DavidCramer 

That's a good point. However, this is an option that isn't exposed to users in the demo, so we do the same. That said, I think this is a value that needs to be added but not necessarily changed.

PR #348, opened 2021-01-25
Exclude AMP pages from Cloudinary URL removal when saving

dugajean:


php/class-cli.php
```diff
@@ -61,7 +61,7 @@ public function __construct( $plugin ) {
 	 * Syncs assets with Cloudinary.
 	 * ## EXAMPLES
 	 *
-	 *     wp cloudinary sync
+	 *     Wp cloudinary sync
```
dugajean, line 64: This change is to satisfy the linter.

php/class-cli.php
```diff
@@ -102,7 +102,7 @@ public function sync( $args, $assoc_args ) {
 	 * Analyze assets with Cloudinary.
 	 * ## EXAMPLES
 	 *
-	 *     wp cloudinary analyze
+	 *     Wp cloudinary analyze
```
dugajean, line 105: This change is to satisfy the linter.

php/class-cli.php
pereirinha: @dugajean 

This is a good catch, but I don't think this should be fixed this way. In systems case sensitive, like Linux, running the `Wp` will generate an error as the command isn't found, so we shouldn't suggest people use it.

A better approach is to define a new rule on `.phpcs.xml.dist`, like this:
```
	<rule ref="Generic.Commenting.DocComment.LongNotCapital">
		<exclude-pattern>php/class-cli.php</exclude-pattern>
	</rule>
```

What do you think?

pereirinha (approved):
@dugajean 

Looks good.
Thank you.

PR #349, opened 2021-01-26
alternate approach

DavidCramer:


php/media/class-video.php
```diff
@@ -95,11 +95,12 @@ public function player_enabled() {
 	 * Initialises the Cloudinary player if it's enabled and if video content is found.
 	 */
 	public function init_player() {
-
 		if ( isset( $this->config['video_player'] ) && 'cld' === $this->config['video_player'] && ! is_admin() ) {
+			global $wp_query;
+			$this->player_enabled = true;
```
pereirinha, line 100: @DavidCramer 

I wonder if it makes sense to set this property here, outside the condition check. Especially because it's set on line 124, inside the condition check.

Thanks.

pereirinha (commented):
@DavidCramer 
 Left you a minor comment.

php/media/class-video.php
DavidCramer: Oh Bother, that's a typo. Sorry, I didn't see it.

pereirinha (approved):
@DavidCramer 

This looks good.

PR #354, opened 2021-01-28
free up memory in CLI

DavidCramer:


pereirinha (approved):
This works great @DavidCramer

PR #355, opened 2021-01-28
Update readme.txt

marissa-masangcay:
Updated readme.txt with updates regarding the deprecation of the v1 plugin. Replaced the "Contributors and Developers" section with this as we already have that listed below.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -53,11 +53,8 @@ And with advanced digital asset management capabilities available in the plugin,
 * [Video transformations documentation](https://cloudinary.com/documentation/video_manipulation_and_delivery?utm_source=wp&utm_medium=wpmarketplace&utm_campaign=wpmarketplace)
 * [FAQ](http://cloudinary.com/faq?utm_source=wp&utm_medium=wpmarketplace&utm_campaign=wpmarketplace)
 
-= Contributors & Developers =
-“Cloudinary – Image and Video Optimization, Manipulation, and Delivery” is an open source software. The following people have contributed to this plugin:
-[Cloudinary](https://profiles.wordpress.org/cloudinary/)
-[XWP](https://profiles.wordpress.org/xwp/)
-[Automattic](https://profiles.wordpress.org/automattic/)
+= Notes =
+The legacy WordPress Plugin version (v1.x) will be deprecated as of February 1st, 2021, after which support, updates and bug fixes for the legacy plugin will continue in limited fashion. The legacy plugin will be made obsolete on August 1st, 2021 (end-of-life date) and will no longer function after that date.
 
 
 == Installation ==


readme.txt
```diff
@@ -53,11 +53,8 @@ And with advanced digital asset management capabilities available in the plugin,
 * [Video transformations documentation](https://cloudinary.com/documentation/video_manipulation_and_delivery?utm_source=wp&utm_medium=wpmarketplace&utm_campaign=wpmarketplace)
 * [FAQ](http://cloudinary.com/faq?utm_source=wp&utm_medium=wpmarketplace&utm_campaign=wpmarketplace)
 
-= Contributors & Developers =
-“Cloudinary – Image and Video Optimization, Manipulation, and Delivery” is an open source software. The following people have contributed to this plugin:
-[Cloudinary](https://profiles.wordpress.org/cloudinary/)
-[XWP](https://profiles.wordpress.org/xwp/)
-[Automattic](https://profiles.wordpress.org/automattic/)
+= Notes =
+The legacy WordPress Plugin version (v1.x) will be deprecated as of February 1st, 2021, after which support, updates and bug fixes for the legacy plugin will continue in limited fashion. The legacy plugin will be made obsolete on August 1st, 2021 (end-of-life date) and will no longer function after that date.
```
asisayag2, line 57: @marissa-masangcay , why is that overriding the `Contributors & Developers` section?

marissa-masangcay: I think I added it in the notes, but we have that mentioned down below in another section with the same title right after 'FAQ' so it's listed twice on the page. It was something I meant to correct a while back so just doing it now.

marissa-masangcay: Unless I misunderstood on the file how it's laid out and should be corrected elsewhere?

asisayag2: Not sure I understand what you tried to fix... Couldn't find anything duplicated

marissa-masangcay: I slacked you a screenshot of what I was referring to a while back. Let me know if you need any further clarification. I may have just not edited the right portion, I'm not exactly sure.

asisayag2: @marissa-masangcay , sorry... I didn't connect the slack to this thread. 
I don't see any other `Contributors & Developers` section in the readme, so I concern that removing this will remove that section all together. 
@pereirinha , wdyt?

pereirinha: @asisayag2 @marissa-masangcay 

I think that Marissa is right. By checking other plugins I've in the WP repo, it seems that the last section on the plugin [landing page](https://wordpress.org/plugins/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/) titled `Contributors & Developers` is auto-generated, meaning that there's no need to have that expressed in the `readme.txt`.

Worst case scenario, if it breaks, we can always add it back :)

gabrielcld2:
Closing this one as the changes due to inactivity and the changes being outdated by now

PR #362, opened 2021-02-12
Bypass syncing of external media

pereirinha:


php/class-media.php
```diff
@@ -244,6 +244,26 @@ public function is_media( $attachment_id ) {
...
+		$is_local_media = false;
+
+		$local_host = wp_parse_url( get_site_url(), PHP_URL_HOST );
+		$media_host = wp_parse_url( get_the_guid( $attachment_id ), PHP_URL_HOST );
+
+		if ( $local_host === $media_host ) {
+			$is_local_media = true;
+		}
+
+		return $is_local_media;
```
DavidCramer, line 264: @pereirinha Wouldn't this be a little simpler?
```suggestion
		return $local_host === $media_host;
```

DavidCramer (approved):

PR #363, opened 2021-02-17
Add support for all supported attachments on WordPress

pereirinha:


php/class-media.php
```diff
@@ -1985,6 +2028,25 @@ public function can_filter_out_local() {
...
+	 * Fix the PDF resource type.
+	 *
+	 * @link https://cloudinary.com/cookbook/convert_pdf_to_jpg
+	 *
+	 * @param string $type          The default type.
+	 * @param int    $attachment_id The attachment ID.
+	 *
+	 * @return string
+	 **/
+	public function pdf_resource_type( $type, $attachment_id ) {
```
DavidCramer, line 2041: @pereirinha - This is a problem. PDF cannot be used as an image, since it will break it as a downloadable. look at 
`Media::is_preview_only()`

DavidCramer (changes requested):
@pereirinha - Generally it looks good. Just concerned with the PDF handling. I think when getting a sized, preview, then its treated as an image, but otherwise, it needs to be handled as a RAW file.
An example is when I inserted a PDF for download, I got the single page, with global transformations applied.

DavidCramer (approved):
@pereirinha - All good, I was incorrect in my test.

PR #364, opened 2021-02-17
Add support logging

DavidCramer:


php/class-plugin.php
```diff
@@ -127,21 +127,21 @@ public function __construct() {
...
+		$this->components['support']      = new Support( $this );
 	}
 
 	/**
 	 * Get a plugin component.
 	 *
 	 * @param mixed $component The component.
 	 *
-	 * @return Connect|Media|REST_API|Settings_Page|Sync|null
+	 * @return Logger|Connect|Media|REST_API|Settings_Page|Sync|null
```
pereirinha, line 144: @DavidCramer 

I think there's a typo here. Isn't that supposed to be `Support`?
I see many other instances of Logger reference, especially on docblocks. Let's fix those as well ;)

php/class-support.php
```diff
@@ -0,0 +1,226 @@
...
+	/**
+	 * Logger constructor.
+	 *
+	 * @param Plugin $plugin Global instance of the main plugin.
+	 */
+	public function __construct( Plugin $plugin ) {
+		parent::__construct( $plugin );
+		add_action( 'cloudinary_settings_save_setting_enable_support', array( $this, 'init_reporting' ), 10, 3 );
+		add_filter( 'media_row_actions', array( $this, 'add_inline_action' ), 10, 2 );
+		add_filter( 'post_row_actions', array( $this, 'add_inline_action' ), 10, 2 );
```
pereirinha, line 39: @DavidCramer 

The option `Add to Cloudinary Report` is not always the last. What do you think if we bump the priority here to ensure that it's always the last?

php/class-support.php
```diff
@@ -0,0 +1,226 @@
...
+		if ( 'attachment' === $post->post_type ) {
+			$meta = wp_get_attachment_metadata( $post->ID );
+
+			$args = array(
+				'type'       => 'tag',
+				'element'    => 'pre',
+				'attributes' => array(
+					'style' => 'overflow:auto;',
+				),
+				'content'    => wp_json_encode( $meta, JSON_PRETTY_PRINT ),
```
pereirinha, line 162: @DavidCramer 

What do you think if we also add the flag for `JSON_UNESCAPED_SLASHES`, as in

```suggestion
				'content'    => wp_json_encode( $meta, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES ),
```

This removes the backslash, and I haven't found any drawbacks.
I tested this in all instances where we output this to the screen, but also on the downloaded file.

php/ui/component/class-system.php
```diff
@@ -0,0 +1,198 @@
...
+
+	/**
+	 * Build the config report.
+	 */
+	protected function config() {
+		$struct['element'] = null;
+
+		$config = $this->setting->get_root_setting()->get_value();
+		unset( $config['connect'] );
+		$config['gallery']['gallery_config'] = json_decode( $config['gallery']['gallery_config'], JSON_PRETTY_PRINT );
```
pereirinha, line 126: @DavidCramer

We need to factor in now the changes we are adding in the other work.

php/ui/component/class-system.php
```diff
@@ -0,0 +1,198 @@
...
+	/**
+	 * Create a report block setting.
+	 *
+	 * @param string $title The title.
+	 * @param string $slug  The slug.
+	 * @param array  $data  The data.
+	 */
+	protected function add_report_block( $title, $slug, $data ) {
+		$this->report_data[ $slug ] = $data;
+		$content                    = wp_strip_all_tags( wp_json_encode( $data, JSON_PRETTY_PRINT ) );
```
pereirinha, line 155: @DavidCramer 

We shouldn't strip the tags here, as this can remove valuable data. I lost the HTML tags that I had inside the content in a classic editor. Let's use `esc_html` instead. Also, consider the `JSON_UNESCAPED_SLASHES`

php/ui/component/class-system.php
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	/**
+	 * Setup the component parts, and build the download report, if needed.
+	 */
+	public function setup_component_parts() {
+		parent::setup_component_parts();
+		$options = array(
+			'options' => function ( $data ) {
+				return json_decode( $data, ARRAY_A );
```
pereirinha, line 179: @DavidCramer 

I'm not sure about the value of `ARRAY_A`, but `json_decode` expects a boolean as a 2nd argument.

Can you double-check the intent here?

php/ui/component/class-system.php
```diff
@@ -0,0 +1,198 @@
...
+			$filename  = "cloudinary-report-{$timestamp}.json";
+			header( 'Content-Description: File Transfer' );
+			header( 'Content-Type: application/octet-stream' );
+			header( "Content-Disposition: attachment; filename={$filename}" );
+			header( 'Content-Transfer-Encoding: text' );
+			header( 'Connection: Keep-Alive' );
+			header( 'Expires: 0' );
+			header( 'Cache-Control: must-revalidate, post-check=0, pre-check=0' );
+			header( 'Pragma: public' );
+			echo wp_json_encode( $data, JSON_PRETTY_PRINT );
```
pereirinha, line 194: Same here about the `JSON_UNESCAPED_SLASHES`.

pereirinha (commented):
@DavidCramer 

This is looking great. I think this will be of great value for us.

Meanwhile, I left you a few comments.

php/class-plugin.php
DavidCramer: @pereirinha - Ah, good catch! That was the original class name, which I changed, but forgot to change the docs.

php/class-support.php
DavidCramer: @pereirinha - Good suggestion!

php/ui/component/class-system.php
DavidCramer: @pereirinha - Good idea!

php/ui/component/class-system.php
DavidCramer: @pereirinha - `ARRAY_A` comes from `wpdb` and is specific to WordPress. I've just been using it for consistency. I actually never thought about it.

PR #367, opened 2021-02-19
Bypass shortcodes with empty args

pereirinha:


DavidCramer (approved):
@pereirinha - perfect.

PR #233, opened 2020-10-13
[WIP] Add old unit tests base, and get working again

DavidCramer:

PR #368, opened 2021-02-20
Add/setting unit tests

DavidCramer:
@pereirinha - #233 is a little complicated approach. This is an alternate, that should be easier to use.


gabrielcld2:
Closing this one due to inactivity and the changes being outdated by now

PR #370, opened 2021-02-22
Fix/video

pereirinha:


php/media/class-filter.php
```diff
@@ -697,6 +697,27 @@ public function catch_media_templates_maybe() {
...
+		if ( 'core/image' === $block['blockName'] ) {
+			remove_filter( 'render_block', array( $this, 'filter_image_block_render_block' ), 10, 2 );
+
+			$filtered_block = $this->filter_image_block_pre_render( $block, $block );
+			$block_content = render_block( $filtered_block );
+
+			add_filter( 'render_block', array( $this, 'filter_image_block_render_block' ), 10, 2 );
+		}
+
+		return $block_content;
```
DavidCramer, line 718: @pereirinha  Instead of adding and removing the filter, you can add a value to the array to looks for and ignore if present. like below:
```suggestion
	public function filter_image_block_render_block( $block_content, array $block ) {
		if ( 'core/image' === $block['blockName'] && empty( $block['cld_render'] ) ) {
			$filtered_block                        = $this->filter_image_block_pre_render( $block, $block );
			$filtered_block['cld_render'] = true;
			$block_content                       = render_block( $filtered_block );
		}

		return $block_content;
```

DavidCramer (approved):
@pereirinha - Looks good to me and is working as expected.
Left you one comment on the adding and removing filters, but it's minor.

PR #373, opened 2021-02-24
Handle other delivery types than upload

pereirinha:


DavidCramer (approved):
@pereirinha - This is cool, no complaints. However I would like us to look at breaking the `Media` class up. It's gotten large and is doing a lot. I'll create a ticket for it and we can chat later.

PR #374, opened 2021-02-24
Ensure that the visibility of controls is defined on settings

pereirinha:


DavidCramer (approved):
@pereirinha - Looks good and is working in my tests.

PR #377, opened 2021-03-01
Ensure that we have gallery settings before trying to use them

pereirinha:


DavidCramer (approved):
@pereirinha looks good. But I'm wanting to make a change in the `get_component`, but let's deal with that later.

PR #382, opened 2021-03-09
Add controls for image crops on PGW

pereirinha:


DavidCramer (approved):
@pereirinha - looks good to me!

js/src/gallery-block/controls.js
```diff
@@ -155,10 +155,23 @@ const Controls = ( { attributes, setAttributes, colors } ) => {
...
+						className="dashicons dashicons-info cld-tooltip"
+						data-tooltip="tooltip_auto_sync"
+					>
+						<span id="tooltip_auto_sync" className="hidden"></span>
+					</span>
+				</p>
 				<p>
 					<ButtonGroup>
-						{ LOOK_AND_FEEL.map( ( type ) => (
+						{ RESIZE_CROP.map( ( type ) => (
```
DavidCramer, line 174: @pereirinha - One thing I learnt from Dugi, Map alteres and returns the altered array, meaning you intend on the array to be changed.
In this case you're just going over each item. So, here the `forEach` would be more appropriate.

DavidCramer (approved):
Other than the minor `map` vs `forEach` All good!

PR #388, opened 2021-03-18
Fix fetched images

pereirinha:


DavidCramer (approved):
@pereirinha - looks good to me.

PR #392, opened 2021-03-22
Sync caption

pereirinha:


php/class-media.php
```diff
@@ -1460,6 +1460,16 @@ private function create_attachment( $asset, $public_id ) {
...
 		}
+		// Capture the Caption Text.
+		if ( ! empty( $asset['meta']['caption'] ) ) {
+			$caption = wp_strip_all_tags( $asset['meta']['caption'] );
+			wp_update_post(
+				array(
+					'ID'           => $asset['attachment_id'],
+					'post_excerpt' => $caption,
+				)
+			);
```
DavidCramer, line 1471: @pereirinha - Do you not think this would be cleaner if you add the `post_excerpt` to the `$post_args` before creating the attachment and keep it in a single call?

pereirinha: Great catch @DavidCramer. You're absolutely right.
I pushed an update.

DavidCramer (approved):

PR #395, opened 2021-03-26
Add option to add report to deactivations and permission for contact

pereirinha:


php/class-deactivation.php
```diff
@@ -205,6 +228,27 @@ public function rest_endpoint( $endpoints ) {
...
+	/**
+	 * Uploads the System Report to the Cloud.
+	 *
+	 * @return array|WP_Error
+	 */
+	public function upload_report() {
+		require_once ABSPATH . '/wp-admin/includes/file.php';
+
+		$report = $this->plugin->get_component( 'report' )->get_report_data();
+		$temp   = get_temp_dir() . $report['filename'];
```
DavidCramer, line 240: @pereirinha - Might be a good idea to use `wp_tempnam()` here, so ensure it's writable and to keep VIP approval.

php/class-deactivation.php
```diff
@@ -230,14 +276,23 @@ public function rest_callback( WP_REST_Request $request ) {
...
+			if ( ! empty( $report['secure_url'] ) ) {
+				$args['report'] = $report['secure_url'];
+			}
+
+			$args['contact'] = $contact;
+		}
+
+		$url = add_query_arg( $args, self::$cld_endpoint );
+
 		$response = wp_remote_get( $url ); // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.wp_remote_get_wp_remote_get
```
DavidCramer, line 296: @pereirinha - I know this is existing code, but I think we should use `wp_safe_remote_get()` here, so we don't need the ignore.

DavidCramer (commented):
@pereirinha Left 2 comments for you

php/class-deactivation.php
pereirinha: @DavidCramer 

As mentioned in our sync, using this function will add a file extension `.tmp`, requiring us to trim it. Otherwise, it will be part of the public_id on Cloudinary. The `wp_tempnam()` is actually a wrapper around `get_temp_dir()` anyways.

php/class-deactivation.php
pereirinha: Thanks, @DavidCramer 

Great catch.

PR #397, opened 2021-03-31
Improve deployment

pereirinha:
This PR will automate creating a `git` tag and a draft for the `GitHub` release.

It introduces 2 new NPM commands:
- `npm run release`;
- `npm run release:dry`;

The first command does the action, whereas the 2nd does the dry run.

The action will also run on the `deploy` command. So, whenever we deploy code to the WP repo, it will also update the GitHub repo.


.release-it.json
```diff
@@ -0,0 +1,17 @@
+{
+  "git": {
+    "requireBranch": "master",
+    "commitMessage": "Version ${version}",
+    "tagName": "${version}",
+    "tagAnnotation": "Version ${version}"
```
kasparsd, line 6: Where is this version coming? Should we move this to a Grunt task?

.release-it.json
```diff
@@ -0,0 +1,17 @@
...
+  },
+  "github": {
+    "release": true,
+    "releaseName": "Version ${version}",
+    "draft": true,
+    "tokenRef": "GITHUB_TOKEN"
```
kasparsd, line 12: Where is this coming from? Should the dev doing the release have this configured?

.release-it.json
pereirinha: Thanks, @kasparsd.

I explored the option of going with the Grunt task — based on the example we saw —, but the tool evolved, and some of the methods are no longer available. 

I've added a new dependency that allows reading the version from a file.
Since we already had logic to update the `readme.txt` in a grunt task, I thought we could use a similar approach to update the `cloudinary.php`.

What do you think?

.release-it.json
pereirinha: Yes, this is an environmental variable that needs to be set in the CI and on the dev machine.
I'm leaving here a [link](https://github.com/release-it/release-it/blob/master/docs/environment-variables.md) to the doc for reference.

pereirinha (commented):
Thanks @kasparsd 

I've pushed some updates that I think will improve the PR.
Happy to discuss these changes in a call.

package.json
```diff
@@ -23,10 +23,13 @@
     "lint:style": "wp-scripts lint-style",
     "lint": "npm-run-all lint:php lint:js lint:style",
     "package": "npm run build && grunt package",
-    "deploy": "npm run build && grunt deploy",
+    "deploy": "npm run build && grunt deploy && npm run release",
```
kasparsd, line 26: Should it tag it before doing the WP.org SVN push? I think the SVN push looks for the latest tag, right? Or do we always push to `trunk` without the tags?

kasparsd (approved):
Looks good!

package.json
pereirinha: We run the replacement of the version in the `package` task which is a dependency of `deploy`. I think that `wp-deploy` uses the plugin version for tagging.

PR #411, opened 2021-04-16
Feature/cache manager

DavidCramer:


pereirinha (commented):
@DavidCramer 

Left you some coments.

php/class-cache.php
```diff
@@ -0,0 +1,1019 @@
...
+	}
+
+	/**
+	 * Register any hooks that this component needs.
+	 */
+	protected function register_hooks() {
+		$this->cache_point = new Cache_Point( $this );
+		add_filter( 'template_include', array( $this, 'frontend_rewrite' ), PHP_INT_MAX );
+		add_action( 'admin_init', array( $this, 'admin_rewrite' ), 0 );
+		add_action( 'deactivate_plugin', array( $this, 'invalidate_plugin_cache' ) );
```
pereirinha, line 319: @DavidCramer 

Is this method implemented?

php/class-cache.php
```diff
@@ -0,0 +1,1019 @@
...
+	 * Register the sync endpoint.
+	 *
+	 * @param array $endpoints The endpoint to add to.
+	 *
+	 * @return mixed
+	 */
+	public function rest_endpoints( $endpoints ) {
+
+		$endpoints['show_cache']          = array(
+			'method'              => \WP_REST_Server::ALLMETHODS,
```
pereirinha, line 333: @DavidCramer 

Why are we exposing all the methods here?

php/class-cache.php
```diff
@@ -0,0 +1,1019 @@
...
+			'callback'            => array( $this, 'rest_get_caches' ),
+			'permission_callback' => array( $this, 'rest_can_manage_options' ),
+			'args'                => array(),
+		);
+		$endpoints['disable_cache_items'] = array(
+			'method'              => \WP_REST_Server::ALLMETHODS,
```
pereirinha, line 339: @DavidCramer 

Same here. I think we should be more especific.

php/class-cache.php
```diff
@@ -0,0 +1,1019 @@
...
+	 */
+	public function sync_static( $file, $url ) {
+
+		$errored = get_transient( self::META_KEYS['upload_error'] );
+		if ( isset( $errored[ $file ] ) && 3 <= $errored[ $file ] ) {
+			unset( $errored[ $file ] );
+			set_transient( self::META_KEYS['upload_error'], $errored, 60 );
+
+			// Dont try again.
+			return new \WP_Error( 'upload_error' );
```
pereirinha, line 449: @DavidCramer 

If we are using `WP_Error`, let's add more details about the error. Perhaps a descriptive message and reference for the errored file. What do you think?

php/class-cache.php
```diff
@@ -0,0 +1,1019 @@
...
+				$upload_file = '@' . $upload_file;
+			}
+		}
+		$options = array(
+			'file'          => $upload_file,
+			'resource_type' => 'auto',
+			'public_id'     => wp_normalize_path( $public_id ),
+		);
+
+		if ( 'image' === $type ) {
```
pereirinha, line 471: @DavidCramer 

Given our sync, this will be optional.

php/class-cache.php
```diff
@@ -0,0 +1,1019 @@
...
+	 * @param string $url  The file URL to upload.
+	 *
+	 * @return string|\WP_Error
+	 */
+	public function sync_static( $file, $url ) {
+
+		$errored = get_transient( self::META_KEYS['upload_error'] );
+		if ( isset( $errored[ $file ] ) && 3 <= $errored[ $file ] ) {
+			unset( $errored[ $file ] );
+			set_transient( self::META_KEYS['upload_error'], $errored, 60 );
```
pereirinha, line 446: @DavidCramer 

I think 1min is short for a new attempt for upload. The current logic will attempt 3 uploads in a minute until it retries in the next minute. In between, we are throwing an error.

See below another instance of this approach, currently on line 482.

php/class-cache.php
```diff
@@ -0,0 +1,1019 @@
...
+	}
+
+	/**
+	 * Check to see if cache setting is enabled.
+	 *
+	 * @param string $cache_setting The setting slug to check.
+	 *
+	 * @return bool
+	 */
+	protected function is_cache_setting_enabled( $cache_setting ) {
```
pereirinha, line 692: @DavidCramer 

I don't think this logic is wired up. I can't disable the cache :)

php/class-cache.php
```diff
@@ -0,0 +1,1019 @@
...
+
+use Cloudinary\Cache\Cache_Point;
+use Cloudinary\Cache\File_System;
+use Cloudinary\Component\Setup;
+use Cloudinary\Settings\Setting;
+
+/**
+ * Plugin report class.
+ */
+class Cache extends Settings_Component implements Setup {
```
pereirinha, line 18: @DavidCramer 

Given the length of this class, what do you think about separating the settings from the rest? It should help to reduce the size and they are isolated concepts. We have already a folder for cache, so we could have a `class-settings.php` inside it.

pereirinha (commented):
@DavidCramer 

I completed reviewing the class cache and left you a few comments.

php/cache/class-cache-point.php
```diff
@@ -0,0 +1,689 @@
...
+	 * @param int       $object_id  The object ID.
+	 * @param string    $meta_key   The meta key.
+	 * @param mixed     $meta_value The meta value.
+	 *
+	 * @return bool
+	 */
+	public function delete_meta( $check, $object_id, $meta_key, $meta_value ) {
+
+		if ( self::POST_TYPE_SLUG === get_post_type( $object_id ) ) {
+			$check = false;
```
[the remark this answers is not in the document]
DavidCramer: @pereirinha - that is what we want. When we hook in, we take over the meta for this post type. it will return false when it doesn't delete. if we returned the null, it will attempt the DB method. we need to have it shorted out.

php/cache/class-cache-point.php
```diff
@@ -0,0 +1,689 @@
...
+		if ( empty( $urls ) ) {
+			return null;
+		}
+
+		$keys        = array_map( array( $this, 'get_key_name' ), $urls );
+		$params      = array(
+			'post_type'      => self::POST_TYPE_SLUG,
+			'post_name__in'  => $keys,
+			'posts_per_page' => - 1,
+			'post_status'    => array( 'draft', 'publish' ),
```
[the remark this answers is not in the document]
DavidCramer: @pereirinha - Yes. If we leave it out, we'll get a mismatch of found and returned, which will trigger the missing items to be recreated.

pereirinha (commented):
@DavidCramer 

I left you some comments. I'll start testing, but I might be circling back to the Cache Point class.

php/cache/class-cache-point.php
```diff
@@ -57,18 +63,52 @@ class Cache_Point {
...
+
 	/**
 	 * Cache Point constructor.
 	 *
 	 * @param Cache $cache The plugin ache object.
 	 */
 	public function __construct( Cache $cache ) {
-		$this->cache = $cache;
+		$this->cache      = $cache;
+		$this->sync_limit = apply_filters( 'cloudinary_on_demand_sync_linmit', 10 );
```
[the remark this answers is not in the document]
DavidCramer: Oops, was meant to do that.

php/cache/class-cache-point.php
```diff
@@ -232,9 +234,9 @@ public function meta_updates() {
 		}
 		// Prep the upload for un-synced items.
 		if ( ! empty( $this->to_upload ) ) {
-			$now = microtime( true );
+			$now = microtime( true ); // Set it further than the initial spawn, to allow it through.
 			wp_schedule_single_event( $now, 'cloudinary_upload_cache', array( $this->to_upload ) );
-			spawn_cron( $now );
+			spawn_cron( $now + 61 );
```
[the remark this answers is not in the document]
DavidCramer: Technically, a minute in seconds would make it === the time. It needs to be higher.
So maybe HOUR_IN_SECONDS 🤔

PR #412, opened 2021-04-16
Fix upgrade assets

pereirinha:


DavidCramer (approved):
@pereirinha - that's a nice simple solution.

PR #419, opened 2021-04-27
add cache status table

DavidCramer:


php/ui/component/class-cache-status.php
```diff
@@ -0,0 +1,75 @@
...
+
+use Cloudinary\Cache;
+use Cloudinary\Cache\Cache_Point;
+
+/**
+ * Cache Status Component to render plan status.
+ *
+ * @package Cloudinary\UI
+ */
+class Cache_Status extends media_status {
```
pereirinha, line 18: @DavidCramer 

Can we update the extended class name to `Media_Status`?

pereirinha (approved):

php/ui/component/class-cache-status.php
DavidCramer: Oops! Will do.

PR #422, opened 2021-04-30
Fix/delete cache item

DavidCramer:


php/cache/class-cache-point.php
```diff
@@ -476,6 +476,48 @@ protected function get_parent_cache_point( $url ) {
...
+		if ( ! is_int( $cache_point_id_url ) ) {
+			$cache_point = $this->get_cache_point( $cache_point_id_url );
+		} else {
+			$cache_point = get_post( $cache_point_id_url );
+		}
+		if ( ! is_null( $cache_point ) ) {
+
+			$params = array(
+				'post_type'              => self::POST_TYPE_SLUG,
+				'posts_per_page'         => 1,
```
pereirinha, line 498: @DavidCramer 

Any reason why we don't do in batches of 100, for instance?

pereirinha (commented):
@DavidCramer 

Left you a minor comment.

php/cache/class-cache-point.php
DavidCramer: Typo! fixing.

pereirinha (approved):
Works great @DavidCramer 
Thanks

PR #425, opened 2021-05-05
Allow to sync not fully synced assets on manual sync

pereirinha:


DavidCramer (approved):
Ah! thats perfect. good catch.

PR #426, opened 2021-05-06
alternate upload via background

DavidCramer:


pereirinha (commented):
@DavidCramer 

I left you a comment.
Also, and following our conversation in the backchannel, let's add the CSS requested SVGs inline. As we know, if they are below a given threshold, they won't be accepted by Cloudinary and we get into the limbo state we saw. Having these inline will also reduce the number of requests.

While in the call we thought about going with the base64 encoding, I'd like you to read this [article](https://css-tricks.com/probably-dont-base64-svg/).

pereirinha (approved):
Looks good.

PR #455, opened 2021-06-16
Fix/sync delete delay

DavidCramer:


pereirinha (approved):
@DavidCramer 

This is working great.
Thank you.

PR #459, opened 2021-06-18
Fix taxonomy transformations

pereirinha:


php/class-delivery.php
```diff
@@ -74,6 +83,36 @@ public function clear_cache() {
...
+			array(
+				'wp-image-',
+				'wp-video-',
+			),
+			array(
+				'wp-post-' . get_the_ID() . ' wp-image-',
+				'wp-post-' . get_the_ID() . ' wp-video-',
+			),
+			$content
+		);
```
DavidCramer, line 104: Nice! I was going to suggest that ;)

php/media/class-global-transformations.php
```diff
@@ -272,8 +273,9 @@ public function get_taxonomy_transformations( $type ) {
 	 * @return bool
 	 */
 	public function is_taxonomy_overwrite() {
-		$apply_type = false;
-		if ( in_the_loop() ) {
+		$apply_type    = false;
+		$maybe_post_id = apply_filters( 'cloudinary_post_id_taxonomy', null );
+		if ( in_the_loop() || $maybe_post_id ) {
 			$post       = get_post();
```
DavidCramer, line 279: @pereirinha - I don't like the repetition here.
Perhaps something like this:
```
		$post = $this->get_current_post();
		if( $post ) { ....       
```
and place the logic in `get_current_post()`

php/media/class-global-transformations.php
```diff
@@ -272,8 +273,9 @@ public function get_taxonomy_transformations( $type ) {
 	 * @return bool
 	 */
 	public function is_taxonomy_overwrite() {
-		$apply_type = false;
-		if ( in_the_loop() ) {
+		$apply_type    = false;
+		$maybe_post_id = apply_filters( 'cloudinary_post_id_taxonomy', null );
```
DavidCramer, line 277: @pereirinha - don't forget to document the new filter 😉

php/media/class-global-transformations.php
```diff
@@ -563,6 +562,29 @@ public function save_overwrite_transformations_featured_image( $post_id ) {
...
+		 * Filter the post ID.
+		 *
+		 * @hook    cloudinary_post_id
+		 * @default null
+		 *
+		 * @return  {WP_Post|null}
+		 */
+		$post_id = apply_filters( 'cloudinary_post_id', null );
+
+		if ( is_null( $post_id ) ) {
```
DavidCramer, line 581: @pereirinha I think get_post returns `null` if outside the loop and no id is provided, making this irrelevant.
However, I'm not concerned and will approve. But if you can, give it a check

DavidCramer (approved):
All good. One comment/question but not important.

DavidCramer (approved):

PR #462, opened 2021-06-22
Fix taxonomy override

pereirinha:


pereirinha:
Closing as it was based on wrong assumptions.

PR #466, opened 2021-06-24
Feature/separate meta

DavidCramer:


php/class-media.php
```diff
@@ -1156,23 +1159,29 @@ public function get_cloudinary_id( $attachment_id ) {
...
+		static $cloudinary_ids = array();
+
+		// Return cached ID if we've already gotten it before.
+		if ( isset( $cloudinary_ids[ $attachment_id ] ) ) {
+			return $cloudinary_ids[ $attachment_id ];
+		}
 
 		if ( ! $this->is_media( $attachment_id ) ) {
+			$cloudinary_ids[ $attachment_id ] = null;
 			return null;
```
pereirinha, line 1174: @DavidCramer 

We are returning `null`. I wonder if this has an impact on the issues we're seeing.

php/class-media.php
```diff
@@ -1878,34 +1887,39 @@ public function build_cached_meta( $post_id, $key, $single ) {
...
-			if ( ! isset( $meta_data[ Sync::META_KEYS['cloudinary'] ] ) ) {
-				$meta_data[ Sync::META_KEYS['cloudinary'] ] = array();
-			}
-			$meta_data[ Sync::META_KEYS['cloudinary'] ][ $key ] = $data;
-			wp_update_attachment_metadata( $post_id, $meta_data );
-		} else {
-			// Update core mete data for consistency.
-			update_post_meta( $post_id, $key, $data );
+
+		$meta = $this->get_post_meta( $post_id );
```
pereirinha, line 1895: @DavidCramer 

I think we have a problem here. Because we don't set the `$single` argument, the return will always be in a wrapping array.

php/class-media.php
```diff
@@ -1878,34 +1887,39 @@ public function build_cached_meta( $post_id, $key, $single ) {
...
+	 * @return bool
 	 */
 	public function delete_post_meta( $post_id, $key ) {
-		$meta_data = wp_get_attachment_metadata( $post_id, true );
-		if ( is_array( $meta_data ) && isset( $meta_data[ Sync::META_KEYS['cloudinary'] ] ) && is_array( $meta_data[ Sync::META_KEYS['cloudinary'] ] ) ) {
-			// Only do this side if has been set before.
-			unset( $meta_data[ Sync::META_KEYS['cloudinary'] ][ $key ] );
-			wp_update_attachment_metadata( $post_id, $meta_data );
+
+		$meta = $this->get_post_meta( $post_id );
```
pereirinha, line 1917: @DavidCramer 

The same here.
I think this is the reason why the meta isn't being deleted.

pereirinha (changes requested):
@DavidCramer 

I left you a few comments.

pereirinha (approved):
Approving after merging the fix branch

PR #469, opened 2021-07-02
restructure asset build

DavidCramer:


pereirinha (approved):
@DavidCramer 

I'm approving this, but I haven't tested too deep.
I'm merging this early so we can "test" while we work on the next release.

PR #473, opened 2021-07-08
Add Cloudinary analytics

pereirinha:


DavidCramer (approved):
Just move after the filter.

PR #474, opened 2021-07-09
Fix/lazy breakpoints cleanup

DavidCramer:


pereirinha (approved):
Thanks, @DavidCramer.
This is working well.

PR #478, opened 2021-07-13
Ensure that attachment_metadata is created and updated

pereirinha:


DavidCramer (approved):
😂 no, not strange. Expected.
Half of me thinks the `wp_generate_attachment_metadata` function should be private with an underscore.

PR #481, opened 2021-07-14
Keep metas that allow getting the ID from given sync keys

pereirinha:


DavidCramer (approved):
I thought this would come back to haunt us. Good catch.

PR #485, opened 2021-07-19
remove suffix version check. clean suffix

DavidCramer:


pereirinha (approved):
Thanks, @DavidCramer 

This is working great.

PR #488, opened 2021-07-26
Versions are supposed to be set by .version

pereirinha:
Prevent being suggested to change on release.


DavidCramer (approved):
Makes sense to me.

PR #494, opened 2021-08-02
Fix build conflict alternative

pereirinha:
@DavidCramer 

I was able to confirm the issue on `develop`. However, looking into the [suggested fix](https://github.com/cloudinary/cloudinary_wordpress/pull/489/), it feels we are re-doing the same definition we do earlier on `sharedConfig.plugins`. 

With this approach, I'm not finding any drawbacks, and the compiled assets are looking smaller.

WDYT?


DavidCramer (approved):
Ye I agree.
That's how I do it in my local. I just didn't get time to update the PR.

Go for it.

PR #504, opened 2021-08-11
Feature/whitelisted domains

pereirinha:


pereirinha:
@DavidCramer 

Many things will change in this PR before it lands on master, but I think this is solid work to be on `develop`. If you agree, feel free to merge it.

PR #505, opened 2021-08-13
rearrange delays, simplify download validation

DavidCramer:


php/connect/class-api.php
```diff
@@ -409,6 +409,8 @@ function_exists( 'wp_get_original_image_url' ) &&
 		// Check if we can try http file upload.
 		if ( ( empty( $headers ) && empty( $disable_https_fetch ) && true === $try_remote ) || ! $media->is_uploadable_media( $attachment_id ) ) {
 			$args['file'] = $file_url;
+		} elseif ( ! $media->is_local_media( $attachment_id ) ) {
```
pereirinha, line 412: @DavidCramer 

I'm still not quite happy here. WDYT if we introduce yet another method called something like `is_external_media`, and the `is_uploadable_media` would be a combination of that and `is_local_media`. This way we can simplify the initial condition to something like `if ( ( empty( $headers ) && empty( $disable_https_fetch ) && true === $try_remote ) || $media->is_external_media( $attachment_id ) ) {`. I suspect that this would be helpful in the future.

pereirinha (approved):
@DavidCramer 

I'm approving and merging.
However, I left you a comment to be considered.

Thanks for taking care of this one.

php/connect/class-api.php
DavidCramer: @pereirinha - I'm actually thinking of splitting the upload method completely to be like, `upload_local` and `upload_remote` then determine which to use at the `upload_asset` level.
I'll do a PR to prototype it.

PR #510, opened 2021-08-31
Fix/new UI settings

DavidCramer:


pereirinha (approved):
Approving and merging for simplicity.
The actual review happens in the baseline branch.

PR #519, opened 2021-09-09
Do not repeat getting the size URLs for the Media Library modal

pereirinha:


DavidCramer (commented):
@pereirinha - I'm torn with this. We use the filter_attachment_for_js to also get the transformations and public id. This change breaks that.
I'm more inclined to simply check that the URL is needed to be converted like here: https://github.com/cloudinary/cloudinary_wordpress/pull/534

PR #526, opened 2021-09-17
add overlays , improve delivery, and improve settings

DavidCramer:


pereirinha (approved):
Thanks, @DavidCramer 

This looks great. I know that there are still a few things off, but I'll merge it as-is for the simplicity of other tasks.

We can address those in a separate PR.

PR #532, opened 2021-09-24
Fix status on large files

pereirinha:


php/class-media.php
```diff
@@ -365,6 +365,34 @@ public function is_uploadable_media( $attachment_id ) {
...
+		}
+
+		$file     = $callback( $attachment_id );
+		$max_size = ( wp_attachment_is_image( $attachment_id ) ? 'image_max_size_bytes' : 'video_max_size_bytes' );
+		if ( file_exists( $file ) && filesize( $file ) > $this->plugin->components['connect']->usage['media_limits'][ $max_size ] ) {
+			$is_oversize = true;
+		}
+
+		return $is_oversize;
+	}
```
DavidCramer, line 394: @pereirinha - since this is hitting the filesystem, and is called many times in a sync check cycle, perhaps add in a static variable cache against the `attachment_id`

php/class-sync.php
```diff
@@ -388,6 +392,10 @@ public function is_syncable( $attachment_id ) {
 			$syncable = true;
 		}
 
+		if ( $this->managers['media']->is_oversize_media( $attachment_id ) ) {
```
DavidCramer, line 395: @pereirinha - Perhaps add this at the end of the chain above, or add `true === $syncable` in front, to stop hitting the file system on files that are not syncable anyway.

php/class-sync.php
```diff
@@ -92,6 +92,7 @@ class Sync implements Setup, Assets {
 		'syncing'        => '_cloudinary_syncing',
 		'downloading'    => '_cloudinary_downloading',
 		'process_log'    => '_process_log',
+		'process_log_v3' => '_cloudinary_process_log',
```
DavidCramer, line 95: @pereirinha - perhaps change the old one to `process_log_v2` or `process_log_legacy` and the new one to just `process_log` this will prevent the version issues when we get to v4.
Although I suspect, this will have implications on where `process_log` is being used elsewhere.

php/class-sync.php
```diff
@@ -198,18 +199,21 @@ public function is_synced( $post_id, $re_check = false ) {
...
-		$log  = $this->managers['media']->get_post_meta( $attachment_id, self::META_KEYS['process_log'], true );
+		$log  = $this->managers['media']->get_process_logs( $attachment_id );
 		$keys = array_keys( $this->sync_base_struct );
 		if ( empty( $log ) || count( $log ) !== count( $keys ) ) {
 			$log = array_fill_keys( $keys, array() );
 		}
 		if ( isset( $log[ $type ] ) ) {
+			if ( is_wp_error( $result ) ) {
+				$result = $result->get_error_message();
+			}
```
DavidCramer, line 210: @pereirinha - To make detecting the result as an error, maybe we make `$result` an array like:
 ```php
 if ( is_wp_error( $result ) ) {
		$result = array(
				'error'   => $result->get_error_code(),
				'message' => $result->get_error_message(),
		);
}
```

**Review by DavidCramer — CHANGES_REQUESTED** [review]
@pereirinha - Thanks for handling this. I left some comments, some things that may help with readability and performance on the file system.
Great work.

File: php/class-media.php [hunk 5]
```diff
@@ -365,6 +366,63 @@ public function is_uploadable_media( $attachment_id ) {
...
+			return $is_oversize[ $attachment_id ];
+		}
+
+		$is_oversize[ $attachment_id ] = false;
+		$file_size                     = $this->get_attachment_file_size( $attachment_id );
+		$max_size                      = ( wp_attachment_is_image( $attachment_id ) ? 'image_max_size_bytes' : 'video_max_size_bytes' );
+		$limit                         = $this->plugin->components['connect']->usage['media_limits'][ $max_size ];
+
+		if ( $file_size > $limit ) {
+			$is_oversize[ $attachment_id ] = true;
```
DavidCramer, line 390: ​We can simplify this with `$​is_oversize​[ ​$​attachment_id​ ]​ = ​$file_size > $limit​;`

PR #582, opened 2021-11-05
Add/event throttling

pereirinha:


DavidCramer (approved):
@pereirinha - I think this is a better way. Good one!

PR #584, opened 2021-11-07
Fix/general fixes

pereirinha:


php/class-media.php
```diff
@@ -1278,7 +1278,11 @@ public function local_url( $attachment_id ) {
 
 		$meta = wp_get_attachment_metadata( $attachment_id );
 		$dirs = wp_get_upload_dir();
-		$url  = wp_normalize_path( trailingslashit( $dirs['baseurl'] ) . $meta['file'] );
+		if ( ! empty( $meta['file'] ) ) {
+			$url = wp_normalize_path( trailingslashit( $dirs['baseurl'] ) . $meta['file'] );
+		} else {
+			$url = get_post_meta( $attachment_id, '_wp_attached_file', true );
+		}
```
DavidCramer, line 1285: @pereirinha - This will only return the relative path if there's no file.
I did correct this in my PR here: https://github.com/cloudinary/cloudinary_wordpress/pull/581/files#diff-16a1b1fbcbe01c2dd4d7849b177e59dda6142990bd71baaceda9870a366a66bdR1280-R1291

DavidCramer (approved):
@pereirinha - All good here, but just check `Media::local_url()` as that has inconsistencies with my PR.

PR #594, opened 2021-11-11
ensure assets are registered and enqueued when media templates are loaded

DavidCramer:


pereirinha (commented):
This was exactly how I was envisioning the fix ;)

pereirinha (approved):
Looks great.
Thank you ;)

PR #598, opened 2021-11-13
Fix/uat cleanups

DavidCramer:


pereirinha (approved):
Thanks for this @DavidCramer

PR #599, opened 2021-11-15
Fix/rc5 fixes

DavidCramer:


DavidCramer:
@pereirinha - Please see #600 which goes with this, but wanted to get your thoughts first.

php/class-settings.php
```diff
@@ -338,35 +338,40 @@ protected function create_child( $slug, $params ) {
...
-		/**
-		 * Filter the setting value.
-		 *
-		 * @hook cloudinary_setting_get_value
-		 *
-		 * @param $value {mixed} The setting value.
-		 * @param $slug  {string}  The setting slug.
-		 */
-		return apply_filters( 'cloudinary_setting_get_value', $value, $slug );
+		return 1 === count( $slugs ) ? array_shift( $return ) : $return;
```
pereirinha, line 374: This took me a while to figure out, but this is very neat to deal with requests where `$slug` is null.

pereirinha (approved):
Great stuff, @DavidCramer 
Thank you.

PR #602, opened 2021-11-16
rework how to find local urls and filter broken CLD only

DavidCramer:


DavidCramer:
creating a better PR - closing.

PR #637, opened 2021-12-20
Fix/database indexes

DavidCramer:


pereirinha (changes requested):
Looks great.

And I agree, let's add the filter.

pereirinha (approved):
LGTM

PR #641, opened 2021-12-22
fix crops on classic

DavidCramer:


pereirinha (approved):
This is working great @DavidCramer.
Thanks.

PR #649, opened 2022-01-04
Remove unneeded sanitization and move sizing concern to the upload

pereirinha:


DavidCramer (approved):
@pereirinha - nice changes. I have questions, but let's not hold it back. Get it out, let's chat tomorrow.

PR #652, opened 2022-01-06
simplify the detection with regex, and check for root, and date folders

DavidCramer:


pereirinha (approved):
The works great.
Thanks

PR #662, opened 2022-01-12
Make sure to include the filename as part of the `public_id` when getting it from URL

pereirinha:


DavidCramer (changes requested):
@pereirinha - This works and solves our immediate problems. However, it doesn't take into account the SEO based urls. We would need to get the `dirname` if it is.

php/class-media.php
```diff
@@ -658,7 +658,12 @@ public function get_public_id_from_url( $url, $as_sync_key = false ) {
 		$file      = implode( '/', $parts );
 		$path_info = pathinfo( $file );
 
-		$public_id = isset( $path_info['dirname'] ) && '.' !== $path_info['dirname'] ? $path_info['dirname'] : $path_info['filename'];
+		// Is SEO friendly URL.
+		if ( 0 === strpos( $path, '/images/' ) ) {
+			$public_id = $path_info['dirname'];
```
DavidCramer, line 663: @pereirinha - Much cleverer!

DavidCramer (approved):

PR #663, opened 2022-01-13
Keep dots as part of the Cloudinary URL on images

pereirinha:


php/connect/class-api.php
```diff
@@ -273,7 +273,7 @@ public function cloudinary_url( $public_id = null, $args = array(), $size = arra
 		$base      = pathinfo( $public_id );
 		// Only do dynamic naming and sizes if upload type.
 		if ( 'image' === $args['resource_type'] && 'upload' === $args['delivery_type'] ) {
-			$new_path  = $base['filename'] . '/' . str_replace( '.', '-', $base['basename'] );
+			$new_path  = $base['filename'] . '/' . $base['basename'];
```
DavidCramer, line 276: @pereirinha - This will still give us an issue if there are dots in the filename, which breaks the URL.
Perhaps the following would be better?
```suggestion
			$clean_base = str_replace( '.', '-', $base['filename'] );
			$new_path   = $base['filename'] . '/' . str_replace( $base['filename'], $clean_base, $base['basename'] );
```

DavidCramer (changes requested):
@pereirinha - Left a comment. Lets sync on this.

DavidCramer (approved):
@pereirinha - Actually, I tested, and it appears the dots in the URL issue, is resolved on Cloudinary's side.
So we're good to go.

PR #664, opened 2022-01-14
add filter to the content temporarily

DavidCramer:


pereirinha (approved):
The solves the problem at hands

PR #666, opened 2022-01-17
move urls to after the tags, append CLDBind

DavidCramer:
this replaces #665


pereirinha (approved):
Looks and works great.
Thank you @DavidCramer

PR #678, opened 2022-02-06
counter inline lazy load script being moved

DavidCramer:


php/delivery/class-lazy-load.php
```diff
@@ -223,7 +227,18 @@ public function enqueue_assets() {
 			$config['placeholder'] = API::generate_transformation_string( $this->get_placeholder_transformations( $config['lazy_placeholder'] ) );
 		}
 		$config['base_url'] = $this->media->base_url;
+
+		ob_start();
```
pereirinha, line 231: @DavidCramer 

Given the fallback in place, is this still needed?

pereirinha (approved):

php/delivery/class-lazy-load.php
DavidCramer: Yes. It's better since we can control it better.

PR #680, opened 2022-02-07
Add checks and relocation methods for non-media assets.

DavidCramer:


pereirinha (approved):
Thanks, @DavidCramer 
This is working great ;)

PR #682, opened 2022-02-08
Fix context sync to include all assets between CLD and WP

DavidCramer:


pereirinha (approved):
This is working great.

PR #683, opened 2022-02-08
Update/plan details

pereirinha:


php/ui/component/class-plan-details.php
```diff
@@ -79,7 +80,7 @@ protected function plan( $struct ) {
 		}
 
 		$struct['children']['requests'] = $this->make_item( __( 'Total Requests', 'cloudinary' ), number_format_i18n( $data['requests'] ), $this->dir_url . 'css/images/requests.svg' );
-		$struct['children']['assets']   = $this->make_item( __( 'Optimized assets', 'cloudinary' ), $data['resources'], $this->dir_url . 'css/images/image.svg' );
+		$struct['children']['assets']   = $this->make_item( __( 'Optimized assets', 'cloudinary' ), number_format_i18n( Sync_Queue::get_optimized_assets() ), $this->dir_url . 'css/images/image.svg' );
```
DavidCramer, line 83: @pereirinha - When the number is 0, it doesn't show since it checks if the content is empty first.
We can resolve this simply by appending a space to the content on line#155 ( Which will also allow maintaining spacing on smaller screens if there is no data)
```
		$description_part['content']               = $description . '&nbsp;';
```
Screenshot 2022-02-09 at 08 36 01
Screenshot 2022-02-09 at 08 30 45

DavidCramer (approved):
@pereirinha - All good. Just that one issue where if there are no assets, it will be blank. But not major or important. Change it if you want.

php/ui/component/class-plan-details.php
DavidCramer: @pereirinha - A better solution is this. The content render has been bugging me a bit, so this PR addresses it better than the space hack above:
https://github.com/cloudinary/cloudinary_wordpress/pull/685

pereirinha: Thanks @DavidCramer

PR #686, opened 2022-02-09
PHPCS config improvements

GaryJones:
- Comment out the `testVersion` property. This doesn't do anything until the `PHPCompatibilityWP` ruleset is pulled in via Composer and added to this PHPCS config.
- Use the PHPCS 3+ format for array value (text domain). The old format is deprecated and scheduled for removal in PHPCS 4.
- Remove `WordPressVIPMinimum` ruleset, as only `WordPress-VIP-Go` should be used. `WordPress-VIP-Go` builds on top of the old `WordPressVIPMinimum` and updates some violation types, severity, and message, so manually including `WordPressVIPMinimum` again was likely undoing those changes.

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <ruleset name="WordPress coding standards">
 	<!-- Check for cross-version support for PHP 5.6 and higher. -->
-	<config name="testVersion" value="5.6-"/>
+	<!--<config name="testVersion" value="5.6-"/>-->
 
 	<config name="ignore_warnings_on_exit" value="1" /><!-- Ignore warnings for now. -->
 
@@ -17,7 +17,9 @@
 
 	<rule ref="WordPress.WP.I18n">
 		<properties>
-			<property name="text_domain" type="array" value="cloudinary" />
+			<property name="text_domain" type="array">
+				<element value="cloudinary"/>
+			</property>
 		</properties>
 	</rule>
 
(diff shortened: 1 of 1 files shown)


phpcs.xml.dist
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <ruleset name="WordPress coding standards">
 	<!-- Check for cross-version support for PHP 5.6 and higher. -->
-	<config name="testVersion" value="5.6-"/>
+	<!--<config name="testVersion" value="5.6-"/>-->
```
pereirinha, line 4: @GaryJones 

Thank you, this is a great catch. However, wouldn't it be beneficial to add the `PHPCompatibilityWP` dependency on the project instead?

pereirinha (commented):
Hi @GaryJones 

First and foremost, thank you for your contribution to the Cloudinary plugin for WordPress.

I left you a question about your thoughts on adding the `PHPCompatibilityWP` instead of commenting out the `testVersion`.

Much appreciated.

phpcs.xml.dist
GaryJones: It would indeed, but that addition of that may not have been desired (for whatever reason), so I was keeping it outside of the scope of this PR. A new PR that adds `PHPCompatibilityWP` ruleset to `composer.json` would also need to edit this file to add the ruleset reference, so could uncomment this config property then.

pereirinha: Thanks for the clarification.

pereirinha (approved):
Thank you once again

PR #694, opened 2022-02-24
decode URL when creating the path so the file can be found

DavidCramer:


pereirinha (approved):
Much better. Thanks @DavidCramer

PR #709, opened 2022-03-09
Replace deprecated `FILTER_SANITIZE_STRING`

pereirinha:


php/templates/taxonomy-transformation-fields.php
```diff
@@ -12,7 +12,7 @@
 
 wp_add_inline_script( 'cloudinary', 'var CLD_GLOBAL_TRANSFORMATIONS = CLD_GLOBAL_TRANSFORMATIONS ? CLD_GLOBAL_TRANSFORMATIONS : {};', 'before' );
 
-$tax_slug   = filter_input( INPUT_GET, 'taxonomy', FILTER_SANITIZE_STRING );
+$tax_slug   = filter_input( INPUT_GET, 'taxonomy', FILTER_CALLBACK, array( 'options' => 'sanitize_text_field' ) );
```
DavidCramer, line 15: @pereirinha - Since we're doing this so often, perhaps it would be a good idea to make it a Util.
```php
	/**
	 * Get a sanitized input text field.
	 *
	 * @param string $var  The value to get.
	 * @param int    $type The type to get.
	 *
	 * @return mixed
	 */
	public static function get_sanitized_text( $var, $type = INPUT_GET ) {
		return filter_input( $type, $var, FILTER_CALLBACK, array( 'options' => 'sanitize_text_field' ) );
	}


$tax_slug = Utils::get_sanitized_text( 'taxonomy' );
```

DavidCramer (changes requested):
@pereirinha - Thanks for this. It's looking way better.  I've added some suggestions.

php/templates/taxonomy-transformation-fields.php
pereirinha: Good idea ;)

php/media/class-global-transformations.php
```diff
@@ -476,11 +476,15 @@ private function init_taxonomy_manager( $post ) {
 	public function save_taxonomy_ordering( $post_id ) {
 		$args = array(
 			'cld_tax_order'  => array(
-				'filter' => FILTER_SANITIZE_STRING,
-				'flags'  => FILTER_REQUIRE_ARRAY,
+				'filter'  => FILTER_CALLBACK,
+				'flags'   => FILTER_REQUIRE_ARRAY,
+				'options' => function( $value ) {
+					return sanitize_text_field( $value );
+				},
```
DavidCramer, line 483: @pereirinha - Couldn't this be the function name, like below?

DavidCramer (approved):
@pereirinha - Looks good! thanks. Left a comment, but up to you if it's important.

PR #738, opened 2022-04-20
include port in clean_url

DavidCramer:
@pereirinha - Please note its against UAT. Needs to be changed to develop once UAT is merged.


pereirinha (approved):
Looks great. Thanks for the catch ;)

PR #755, opened 2022-05-23
strip out saved transformations

DavidCramer:


pereirinha (approved):
This is working great.

PR #756, opened 2022-05-24
Feature/improve support metadata

pereirinha:
Adds support to replacement of Cloudinary URLs, when storing metadata.

### DIFF
--- a/php/class-delivery.php
+++ b/php/class-delivery.php
@@ -154,6 +154,49 @@ class Delivery implements Setup {
 	}
 
 	/**
+	 * Maybe filter out Cloudinary URLs in post meta.
+	 *
+	 * @param null|bool $check      Whether to allow adding metadata for the given type.
+	 * @param int       $object_id  The ID of the object metadata is for.
+	 * @param string    $meta_key   The Metadata key.
+	 * @param mixed     $meta_value Metadata value.
+	 *
+	 * @return null|bool
+	 */
+	public function maybe_filter_out_metadata( $check, $object_id, $meta_key, $meta_value ) {
+
+		$current       = current_filter();
+		$is_serialized = false;
+		$is_object     = is_object( $meta_value );
+
+		if ( ! is_string( $meta_value ) ) {
+			$is_serialized = true;
+			$meta_value    = wp_json_encode( $meta_value );
+		}
+
+		list( $action, $object ) = explode( '_', $current );
+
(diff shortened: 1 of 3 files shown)


php/class-delivery.php
```diff
@@ -153,6 +153,49 @@ function ( $post ) {
...
+	 */
+	public function maybe_filter_out_metadata( $check, $object_id, $meta_key, $meta_value ) {
+
+		$current       = current_filter();
+		$is_serialized = false;
+		$is_object     = is_object( $meta_value );
+
+		if ( ! is_string( $meta_value ) ) {
+			$is_serialized = true;
+			$meta_value    = wp_json_encode( $meta_value );
```
DavidCramer, line 174: @pereirinha  - I think we should have a check here to be sure that it was encoded.

php/class-delivery.php
```diff
@@ -153,6 +153,49 @@ function ( $post ) {
...
+			if ( $is_serialized ) {
+				$meta_value = json_decode( wp_unslash( $process_meta_value ), ! $is_object );
+			} else {
+				$meta_value = $process_meta_value;
+			}
+			$check = call_user_func( "{$action}_{$object}_meta", $object_id, $meta_key, $meta_value );
+		}
+
+		add_filter( $current, array( $this, 'maybe_filter_out_metadata' ), 10, 4 );
+
```
DavidCramer, line 195: @pereirinha - instead of removing and adding the filter, perhaps we could use a static variable. Set it true when processing, then false after. And return immediately if its true.

DavidCramer, line 195: @pereirinha - instead of removing and adding the filter, perhaps we could use a static variable. Set it true when processing, then false after. And return immediately if its true.

DavidCramer (approved):
@pereirinha - Looks good and works as I would expect. I left minor suggestions.

php/class-delivery.php
pereirinha: This is a great point. Thanks. I've updated the code.

php/class-delivery.php
pereirinha: I don't have a strong opinion on this one. By default, I'd go with the adding and removing of filters, as it makes the reading easier of what you're trying to achieve, but I see that we use flags in other instances as well, so, for consistency, I've updated the code.

DavidCramer (approved):

PR #773, opened 2022-07-06
move transformations to relationship table

DavidCramer:
Move the transformations to relationship table.

### DIFF
--- a/php/class-relate.php
+++ b/php/class-relate.php
@@ -70,10 +70,6 @@ class Relate {
 	public function upgrade_relation( $attachment_id, $version ) {
 		$asset_plugin_version = $this->media->get_post_meta( $attachment_id, Sync::META_KEYS['plugin_version'], true );
 		if ( ! empty( $asset_plugin_version ) && version_compare( $asset_plugin_version, $version, '<' ) ) {
-			$transformations = $this->media->get_post_meta( $attachment_id, Sync::META_KEYS['transformation'], true );
-			if ( ! empty( $transformations ) ) {
-				self::update_transformations( $attachment_id, $transformations );
-			}
 			$this->media->delete_post_meta( $attachment_id, Sync::META_KEYS['transformation'] );
 		}
 	}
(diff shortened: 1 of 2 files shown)


php/class-relate.php
```diff
@@ -0,0 +1,118 @@
...
+	/**
+	 * Update transformations for an asset.
+	 *
+	 * @param int          $attachment_id   The attachment ID.
+	 * @param array|string $transformations The transformations.
+	 */
+	public static function update_transformations( $attachment_id, $transformations ) {
+		$relationship = Relationship::get_relationship( $attachment_id );
+		if ( is_array( $transformations ) ) {
+			$transformations = Api::generate_transformation_string( $transformations, $relationship->asset_type );
```
pereirinha, line 90: @DavidCramer 

I'm not sure we want to do this.

@Vdeub-cloudinary

With this suggested change, we'll override everything currently in the relationship table with the transformations within the post meta. This might revert changes done manually in the relationship table.

Do you have any concerns here?

pereirinha (commented):
@DavidCramer 

I've found something that I'd like @Vdeub-cloudinary to comment on.

php/class-relate.php
DavidCramer: we really do. 
the relationship table its stored ad a string, in meta its stored as an array.
In the the get transformations method, i've added the flag to deliver in a string or an array.

php/class-relate.php
```diff
@@ -0,0 +1,118 @@
...
+	/**
+	 * Upgrade an asset relation.
+	 *
+	 * @param int    $attachment_id The attachment ID.
+	 * @param string $version       The version upgrading to.
+	 */
+	public function upgrade_relation( $attachment_id, $version ) {
+		$asset_plugin_version = $this->media->get_post_meta( $attachment_id, Sync::META_KEYS['plugin_version'], true );
+		if ( ! empty( $asset_plugin_version ) && version_compare( $asset_plugin_version, $version, '<' ) ) {
+			$transformations = $this->media->get_post_meta( $attachment_id, Sync::META_KEYS['transformation'], true );
```
pereirinha, line 73: @DavidCramer 

I'm sorry, I should had comment here instead of https://github.com/cloudinary/cloudinary_wordpress/pull/773#discussion_r916072103

pereirinha (approved):
Thanks @DavidCramer 

This is working great.

PR #777, opened 2022-07-11
disallow dots in file name

DavidCramer:
Fixes an issue that loses the setting point if the slug has a dot present.


pereirinha (approved):
Seems to be working as expected.

PR #779, opened 2022-07-12
Check get_current_screen exists

DavidCramer:
check for `get_current_screen` before trying to use it.


pereirinha (approved):
Nice and simple

PR #780, opened 2022-07-13
Add dynamic inline script loader

DavidCramer:
Dynamically add the inline loader if the content is loaded via an endpoint.


pereirinha (commented):
@DavidCramer 

I left you some comments. 
I don't think that there are blockers, but I'd like to get your feedback.

pereirinha (approved):
**👏 praise:** This is great. Thanks

PR #781, opened 2022-07-13
Make use of `Utils::user_can()`

pereirinha:
Puts all user permissions checks behind a Cloudinary filter

### DIFF
--- a/php/assets/class-rest-assets.php
+++ b/php/assets/class-rest-assets.php
@@ -53,12 +53,12 @@ class Rest_Assets {
 		$endpoints['show_cache']          = array(
 			'method'              => \WP_REST_Server::CREATABLE,
 			'callback'            => array( $this, 'rest_get_caches' ),
-			'permission_callback' => array( $this, 'rest_can_manage_options' ),
+			'permission_callback' => array( $this, 'rest_can_manage_assets' ),
 			'args'                => array(),
 		);
 		$endpoints['disable_cache_items'] = array(
 			'method'              => \WP_REST_Server::CREATABLE,
-			'permission_callback' => array( $this, 'rest_can_manage_options' ),
+			'permission_callback' => array( $this, 'rest_can_manage_assets' ),
 			'callback'            => array( $this, 'rest_handle_state' ),
 			'args'                => array(
 				'ids'   => array(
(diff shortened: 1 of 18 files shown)


php/class-media-library.php
```diff
@@ -41,7 +41,7 @@ public function setup() {
 		$this->handle = add_menu_page(
 			__( 'Cloudinary Media Library', 'cloudinary' ),
 			__( 'Cloudinary DAM', 'cloudinary' ),
-			'manage_options',
+			Utils::user_can( 'manage_dam' ),
```
DavidCramer, line 44: This will result in passing a boolean into the capability param for the `add_menu_page()` adding a `true` will result in the `map_meta_cap()` defaulting to `remove_users` Which will allow admins to view it.
But if `manage_dam` is filtered to `upload_files` so that editors can use it, it will not work.

Perhaps something like this would be better?
```suggestion
			Utils::user_can( 'manage_dam' ) ? 'exist' : false,
```

php/class-plugin.php
```diff
@@ -190,7 +190,7 @@ private function get_settings_page_structure() {
 			'version'    => $this->version,
 			'page_title' => __( 'Cloudinary', 'cloudinary' ),
 			'menu_title' => __( 'Cloudinary', 'cloudinary' ),
-			'capability' => 'manage_options',
+			'capability' => Utils::user_can( 'manage_settings' ),
```
DavidCramer, line 193: Same as with the media library, will result in only admins using.
suggest:
```suggestion
			'capability' => Utils::user_can( 'manage_settings' )  ? 'exist' : false,
```

php/class-rest-api.php
```diff
@@ -67,7 +67,7 @@ public function rest_api_init() {
 	 * @return bool
 	 */
 	public static function rest_can_manage_options() {
```
DavidCramer, line 69: Also maybe rename method?

DavidCramer (changes requested):
@pereirinha - Thanks for this, it's great and offers a more controlled way to manage access for users.
There are however a few concerns, that needs to be looked at.

php/assets/class-rest-assets.php
```diff
@@ -239,7 +239,7 @@ public function rest_get_caches( $request ) {
 	 * @return bool
 	 */
 	public function rest_can_manage_options() {
```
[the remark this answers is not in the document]
pereirinha: Good idea ;)

pereirinha:
@DavidCramer 

Ready for you to circle back.

PR #782, opened 2022-07-15
Support spaces folder name

pereirinha:
When the Cloudinary sync folder contains spaces in its name, the filter out of Cloudinary URLs fails to detect the full URL for the image, corrupting the data stored in the DB.

### DIFF
--- a/php/class-delivery.php
+++ b/php/class-delivery.php
@@ -242,6 +242,8 @@ class Delivery implements Setup {
 				$working_content = $unslash_maybe;
 			}
 		}
+		// Sometimes, synced folders contains spaces. For properly extract URLs, we need to replace the spaces with %20.
+		$working_content = str_replace( ' ', '%20', $working_content );
 		$base_urls       = array_unique( Utils::extract_urls( $working_content ) );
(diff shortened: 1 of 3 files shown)


php/class-delivery.php
```diff
@@ -242,6 +242,8 @@ public function filter_out_cloudinary( $content ) {
 				$working_content = $unslash_maybe;
 			}
 		}
+		// Sometimes, synced folders contains spaces. For properly extract URLs, we need to replace the spaces with %20.
+		$working_content = str_replace( ' ', '%20', $working_content );
```
DavidCramer, line 246: @pereirinha - Maybe using `esc_url()` would be better here?

DavidCramer (approved):
all good, just a comment on function use.

DavidCramer:
@pereirinha - Perhaps, instead of removing spaces in the working content, we should simply `esc_url()` in the `Media::get_cloudinary_folder()` here: https://github.com/cloudinary/cloudinary_wordpress/blob/uat/php/class-media.php#L1467

DavidCramer (changes requested):
@pereirinha This is cool. A much simpler solution. I left a comment about caching, as well as a PR against this.

PR #792, opened 2022-07-27
Fix filter out fetch images

pereirinha:
Fixes the replacement of Cloudinary URLs for fetched images.

### DIFF
--- a/php/class-media.php
+++ b/php/class-media.php
@@ -650,6 +650,7 @@ class Media extends Settings_Component implements Setup {
 		$parts = explode( '/', ltrim( $path, '/' ) );
 
 		$maybe_seo = array();
+		$public_id = null;
 
 		// Need to find the version part as anything after this is the public id.
 		foreach ( $parts as $part ) {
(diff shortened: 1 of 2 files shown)


php/class-media.php
```diff
@@ -671,6 +672,17 @@ public function get_public_id_from_url( $url, $as_sync_key = false ) {
 		// Is SEO friendly URL.
 		if ( in_array( 'images', $maybe_seo, true ) ) {
 			$public_id = $path_info['dirname'];
+		} elseif ( false !== strpos( $url, '/image/fetch/' ) ) {
+			// Maybe the $file is already the URL - $url has the version.
+			if ( filter_var( $file, FILTER_VALIDATE_URL ) ) {
```
DavidCramer, line 677: @pereirinha perhaps utf8 encode the url in case of international chars?

DavidCramer (approved):
All good, just one comment.

php/class-media.php
pereirinha: @DavidCramer, please check the comment about this is the backlog.

Issue #791, opened 2022-07-27
PHPCompatibiltyWP missing from PHPCS config file


GaryJones:
## Bug Description

A `testVersion` configuration value in the PHPCS config file was [uncommented](https://github.com/cloudinary/cloudinary_wordpress/commit/d8bc42264d4179d00e44a56669faccf265e3b5db) by @pereirinha, but it still doesn't have any effect until the PHPCompatibility or PHPCompatibilityWP rule is added. The line was previously [commented out](https://github.com/cloudinary/cloudinary_wordpress/commit/19064d2ff8c57c51e9fdde059649e7562c159d87) because it wasn't having any effect then. 

## Expected Behaviour

The [PHPCS config](https://github.com/cloudinary/cloudinary_wordpress/blob/master/phpcs.xml.dist) file should contain something like:

```xml
<!-- Rules: Check PHP version compatibility - see
		https://github.com/PHPCompatibility/PHPCompatibilityWP -->
<rule ref="PHPCompatibilityWP"/>
```

## Steps to reproduce

[PHPCompatibilityWP has been added as a dev dependency](https://github.com/cloudinary/cloudinary_wordpress/commit/2a61cf1df75583e3af2f3ef405580e910cc668e5) (good), but it isn't in the [PHPCS config](https://github.com/cloudinary/cloudinary_wordpress/blob/master/phpcs.xml.dist) where it would be expected to be found along with the other ruleset references.

See https://github.com/Parsely/wp-parsely/blob/d0384362e1815aa557a5447b738d8bac58dbcb80/.phpcs.xml.dist#L30-L35 as an example.

GaryJones:
Separately, I'll also note that you have the `*` constraint for the `PHPCompatibilityWP` package. This will also pull in  `PHPCompatibility` v9, but that doesn't have the sniffs related to PHP 8.

Please follow the steps at https://docs.wpvip.com/how-tos/code-scanning-for-php-upgrade/#Upcoming-releases-of-PHPCompatibility to use `PHPCompatibilityWP` with `PHPCompatibility` `develop` branch which does contain the PHP 8-related sniffs.

PixelCook:
Gary I've submitted your notes to our production team. Thanks so much! I'll update with any updates I have.

pereirinha:
Thanks, @GaryJones, for raising this.

[...]

PR #794, opened 2022-07-28
Fix phpcs config file

pereirinha:
Addresses https://github.com/cloudinary/cloudinary_wordpress/issues/791

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
     "wp-coding-standards/wpcs": "^2.3",
     "automattic/vipwpcs": "^2.2",
     "dealerdirect/phpcodesniffer-composer-installer": "^0.7.0",
-    "phpcompatibility/phpcompatibility-wp": "*"
+    "phpcompatibility/phpcompatibility-wp": "^2.1"
   },
   "config": {
     "platform": {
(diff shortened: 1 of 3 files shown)


GaryJones (approved):
In terms of adding PHPCompatibilityWP, this is ✅.

However, note that the current release of PHPCompatibilityWP uses a version of PHPCompatibility that does NOT include the PHP 8-related sniffs, which are only in the `develop` branch of `PHPCompatibility`, so [further tweaks](https://docs.wpvip.com/how-tos/code-scanning-for-php-upgrade/#Upcoming-releases-of-PHPCompatibility) would be needed if checking for PHP 8 compatibility is what's desired.

pereirinha:
Thanks again, @GaryJones, for pointing out further improvements.
I did update both PRs, so they should now fully cover the upcoming PHPCompatibility 10.

I'll keep an eye on it so that when it's out, we can update our code base accordingly.

Can you please do just a sanity check here to validate ahead of merging it? Much appreciated.

GaryJones (approved):
Looks good to me!

pereirinha:
Thank you @GaryJones.
Merging.

PR #798, opened 2022-08-02
filter result of REST request

DavidCramer:
Move filtering the result of an api request to pre echo. (exclude internal requests)


pereirinha (approved):
Looks good

PR #799, opened 2022-08-08
Fix/convert incompatible types

DavidCramer:
fix the issue where a url can't be checked for file type.


pereirinha (approved):
Works great

PR #801, opened 2022-08-22
Introduce wrapper and address recent introduced methods

pereirinha:
Fixes the plugin in older versions of WordPress.


DavidCramer (approved):
@pereirinha - Good catches here. Forget to check that @since tags 🤦

PR #803, opened 2022-09-13
Enqueue Cloudinary script for the Editor

pereirinha:
Fixes broken terms order

PR #804, opened 2022-09-13
Fix/terms order

pereirinha:
Extracts the terms order script and enqueues it on the Editor screen, as an alternative approach to #803


DavidCramer (commented):
@pereirinha - I prefer this approach, however it also requires more testing. Let's chat about it on our monday sync.

DavidCramer (approved):
@pereirinha - I'm happy with this. Please see #813

PR #808, opened 2022-09-19
Partial support for above the fold lazy load

pereirinha:


DavidCramer (approved):
@pereirinha - Nice work. I played with this a bit and tested various scenarios and were good.
I think there's a little more we can still do, but this is a good start to get feedback from.

PR #809, opened 2022-09-20
Improves consistency

pereirinha:


DavidCramer (approved):
@pereirinha - good catch.

PR #810, opened 2022-09-23
Fix php 8.0 and 8.1 warnings

pereirinha:


php/delivery/class-lazy-load.php
```diff
@@ -57,7 +57,7 @@ public function setup_hooks() {
 	 */
 	public function init_lazy_script_maybe() {
 
-		$flag = filter_input( INPUT_GET, 'cloudinary_lazy_load_loader', FILTER_SANITIZE_STRING );
+		$flag = ! empty( $_GET['cloudinary_lazy_load_loader'] ) ? htmlspecialchars( $_GET['cloudinary_lazy_load_loader'] ) : null;
```
DavidCramer, line 60: @pereirinha : should we not be using the `Utils::get_sanitized_text()` for this purpose?
https://github.com/cloudinary/cloudinary_wordpress/blob/d9263d464db608976a44681f09ec2948943e274b/php/class-utils.php#L474-L484

DavidCramer (changes requested):
@pereirinha - added a question.

php/delivery/class-lazy-load.php
DavidCramer: @pereirinha - Maybe we can alter that and add an ability to change the callback?

DavidCramer (approved):

PR #811, opened 2022-09-23
SVGs shouldn't work with responsive image

pereirinha:


DavidCramer (approved):
@pereirinha - Nice, I like that you added the filter as well.

PR #813, opened 2022-09-29
fix delay

DavidCramer:
The terms can take a bit of time to get populated, and the subscribe fails.


DavidCramer:
Adds an additional fix for #804

pereirinha (approved):
LGTM

PR #816, opened 2022-10-07
Feature disable sync per type

pereirinha:


DavidCramer (approved):
@pereirinha - Nothing really stands out but it's a big PR. It appears to be working, so approving now, but we need to make sure we test the package a lot.

PR #817, opened 2022-10-11
Fix video format

pereirinha:


php/media/class-video.php
```diff
@@ -311,21 +311,14 @@ protected function build_video_embed( $attachment_id, $attributes = array(), $ov
...
 			),
-			'source'     => array(
-				'source_types' => array(),
-			),
 		);
 		// Check for transformations.
 		$transformations = $this->media->get_transformations( $attachment_id, array(), $overwrite_transformations );
 		if ( ! empty( $transformations ) ) {
-			$params['source']['transformation'] = $transformations;
+			$params['source']['transformation'] = array_merge( ...$transformations );
```
DavidCramer, line 318: @pereirinha - by having all the transformations in a single band effects won't work. for example `e_reverse/e_progressbar:bar:green` the second effect will be dropped.

DavidCramer (changes requested):
@pereirinha - the issue here is with the embed script and not something we can fix. Lets chat about this in our sync.

PR #844, opened 2023-01-09
Get a keyless array when getting the size from URL

pereirinha:


DavidCramer (approved):
@pereirinha - Nice catch.

PR #848, opened 2023-01-20
Fix/cron

DavidCramer:
Revise starting time to prevent impossible numbers - shorter run times, longer wait cycle


pereirinha (approved):
Looks good.

PR #852, opened 2023-01-20
Hotfix/asset delivery

pereirinha:


ui-definitions/settings-video.php
```diff
@@ -35,13 +37,23 @@
...
 						'cloudinary'
 					),
-					'description'        => __( 'Optimize and deliver videos on my site.', 'cloudinary' ),
+					'description'        => __( 'Sync and deliver videos from Cloudinary.', 'cloudinary' ),
 					'default'            => 'on',
 					'attributes'         => array(
-						'data-context' => 'image',
+						'data-context' => 'video',
+					),
+					'disabled'           => ! get_plugin_instance()->get_component( 'storage' )->is_local_full(),
```
DavidCramer, line 48: @pereirinha - If you place this in a callback, then in the component, you call it then, the settings API will be available.

DavidCramer (approved):
@pereirinha - This is cool. Just left a single suggestion. Can use it or not.

ui-definitions/settings-video.php
pereirinha: @DavidCramer 

I moved the approach to the suggested one.
Mind circling back?

PR #853, opened 2023-01-25
Check against offload key

pereirinha:


php/sync/class-storage.php
```diff
@@ -532,6 +532,6 @@ public function setup() {
 	 * @return bool
 	 */
 	public function is_local_full() {
-		return 'dual_full' === $this->media->get_settings()->get_value( 'sync_media' );
+		return 'dual_full' === $this->media->get_settings()->get_value( 'sync_media' )['offload'];
```
DavidCramer, line 535: @pereirinha I would assign `$this->media->get_settings()->get_value( 'sync_media' )` to a variable, then check if offload isset and ===. Just to be sure.

DavidCramer (changes requested):
@pereirinha slight change.

DavidCramer (approved):
@pereirinha - Much better :)

PR #854, opened 2023-01-27
Do not escape the URL. We built it

pereirinha:


DavidCramer (approved):
@pereirinha - good catch

PR #859, opened 2023-02-09
Fix import assets without delivery

pereirinha:


DavidCramer (approved):
@pereirinha - It appears to be working as expected.

PR #875, opened 2023-04-11
Allow saving empty string on saving term freeform transformation

pereirinha:


DavidCramer (approved):
@pereirinha Good catch.

PR #893, opened 2023-07-31
Fix WPVIP feedback

pereirinha:
This PR addresses some warnings seen on WPVIP.

Aside from minor spacing fixes and alike, it also introduces the following:
- `Utils::attachment_url_to_postid` — a wrapper for `wpcom_vip_attachment_url_to_postid` or use `wp_cache` to store this information. This should have a positive impact on performance;
- fixes on `export_csv` for the CLI commands, which due to the VIP filesystem, could result in an impossibility of storing the export data;
- Increases the sensibility of the PHPCS so that we can sooner spot these issues;

### DIFF
--- a/php/class-utils.php
+++ b/php/class-utils.php
@@ -923,14 +923,17 @@ class Utils {
 	 * @return int
 	 */
 	public static function attachment_url_to_postid( $url ) {
-		$attachment_id = 0;
+		$key = "postid_{$url}";
+
 		if ( function_exists( 'wpcom_vip_attachment_url_to_postid' ) ) {
 			$attachment_id = wpcom_vip_attachment_url_to_postid( $url );
+		} else {
+			$attachment_id = wp_cache_get( $key, 'cloudinary' );
 		}
 
-		if ( 0 === $attachment_id && empty( wp_cache_get( "postid_{$url}", 'cloudinary' ) ) ) {
+		if ( empty( $attachment_id ) ) {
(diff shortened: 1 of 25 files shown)


php/class-utils.php
```diff
@@ -914,4 +914,25 @@ public static function get_registered_sizes( $attachment_id = null ) {
...
+	public static function attachment_url_to_postid( $url ) {
+		$attachment_id = 0;
+		if ( function_exists( 'wpcom_vip_attachment_url_to_postid' ) ) {
+			$attachment_id = wpcom_vip_attachment_url_to_postid( $url );
+		}
+
+		if ( 0 === $attachment_id && empty( wp_cache_get( "postid_{$url}", 'cloudinary' ) ) ) {
+			$attachment_id = attachment_url_to_postid( $url ); // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.attachment_url_to_postid_attachment_url_to_postid
+			wp_cache_set( "postid_{$url}", $attachment_id, 'cloudinary' );
+		}
```
DavidCramer, line 934: @pereirinha - You're not actually using the value from wp_cache_get. Perhaps this would be better placed when first defining the var: `$attachment_id = wp_cache_get( "postid_{$url}", 'cloudinary' );`
Then wrap all the logic in if ! empty

DavidCramer (changes requested):
@pereirinha - It's all cool, but the logic in `attachment_url_to_postid` is not using the cache.

php/class-utils.php
pereirinha: @DavidCramer 

Great catch. Ready for you to circle back ;)

DavidCramer (approved):
@pereirinha - Great stuff! Back to you.

PR #898, opened 2023-08-22
Fix protected property

pereirinha:


DavidCramer (approved):
All good @pereirinha .

PR #910, opened 2023-10-25
Add adaptive streaming

DavidCramer:
Deliver video in  .m3u8 and .mpd formats.


php/class-media.php
```diff
@@ -1230,11 +1230,25 @@ public function apply_default_transformations( array $transformations, $attachme
...
 			'qf'     => array(),
 		);
+		// Adaptive streaming eagers.
+		if ( 'video' === $type ) {
+			$streaming = $this->settings->get_value( 'adaptive_streaming', 'adaptive_streaming_mode' );
+			if ( 'on' === $streaming['adaptive_streaming'] ) {
+				$streaming_transform          = array(
+					array(
+						'streaming_profile' => 'auto',
+						'fetch_format'      => $streaming['adaptive_streaming_mode'],
```
pereirinha, line 1246: **❗️issue:** 

@DavidCramer I think this is where I'm getting an error when syncing new videos, or trying to play videos that have been published before this code base.

The error is:
`streaming_profile must be the only directive in the transformation component` 

Can you please have a look here?

pereirinha (changes requested):
@DavidCramer 

I left you a couple of comments.

php/class-media.php
yuval-cloudinary: fetch_format should not be added to the transformation, but rather the URL extension should be changed - .m3ud for HLS and .mpd for MPEG-DASH

pereirinha (approved):
Ok, let's merge as is an test with the other work.

PR #912, opened 2023-10-31
Videos need the format as part of the URL

pereirinha:


DavidCramer (approved):
@pereirinha - All good. Does not cause issues with the other video ticket.

PR #914, opened 2023-11-10
Fix storage stats

pereirinha:


DavidCramer (approved):
@pereirinha - All good!

PR #918, opened 2023-12-18
Update stats with Cron Job

pereirinha:


DavidCramer (commented):
@pereirinha - Looks good to me. My only thought it to be sure that every minute is not too short a time. We need to be sure that it doesn't take longer than a minute per cycle.

DavidCramer (approved):

PR #922, opened 2024-01-09
Introduce filter to control the `content_url`. Useful when the FE dom…

pereirinha:


DavidCramer (approved):
@pereirinha - Looks good to me!

PR #927, opened 2024-01-17
add additional timeout for each inline transformation

DavidCramer:
Multiplies the timeout based on how many inline transformations are added. This will account for the additional eagers, which takes time.


pereirinha (approved):
Great work @DavidCramer 

Works like a charm.

PR #929, opened 2024-01-24
Fix moved method

pereirinha:


DavidCramer (approved):
@pereirinha Nice. Looks good to me.

PR #955, opened 2024-06-07
Feature WPML support

pereirinha:
Add support for WPML

### DIFF
--- a/php/class-assets.php
+++ b/php/class-assets.php
@@ -1020,7 +1020,8 @@ class Assets extends Settings_Component {
 		global $wpdb;
 
 		$wpdb->cld_table = Utils::get_relationship_table();
-		$prepare         = $wpdb->prepare( "SELECT * FROM $wpdb->cld_table WHERE post_id = %d;", (int) $post_id );
+		$media_context   = Utils::get_media_context( $post_id );
+		$prepare         = $wpdb->prepare( "SELECT * FROM $wpdb->cld_table WHERE post_id = %d AND media_context = %s;", (int) $post_id, $media_context );
 		$result          = $wpdb->get_row( $prepare, ARRAY_A ); // phpcs:ignore WordPress.DB.DirectDatabaseQuery, WordPress.DB.PreparedSQL.NotPrepared
 
 		return $this->build_item( $result );
--- a/php/class-utils.php
+++ b/php/class-utils.php
@@ -361,17 +361,27 @@ class Utils {
 	 * @return array
 	 */
 	protected static function get_upgrade_sequence() {
-		$sequence         = array();
+		$upgrade_sequence = array();
 		$sequences        = array(
-			'3.0.0' => array( 'Cloudinary\Utils', 'upgrade_3_0_1' ),
+			'3.0.0' => array(
+				'range'  => array( '3.0.0' ),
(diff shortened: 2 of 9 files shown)


php/class-assets.php
```diff
@@ -1020,7 +1020,8 @@ public function get_asset( $post_id, $type = 'object' ) {
 		global $wpdb;
 
 		$wpdb->cld_table = Utils::get_relationship_table();
-		$prepare         = $wpdb->prepare( "SELECT * FROM $wpdb->cld_table WHERE post_id = %d;", (int) $post_id );
+		$media_context   = Utils::get_media_context( $post_id );
+		$prepare         = $wpdb->prepare( "SELECT * FROM $wpdb->cld_table WHERE post_id = %d AND media_context = %s;", (int) $post_id, $media_context );
```
yuval-cloudinary, line 1024: Should we add media_context to the query in https://github.com/cloudinary/cloudinary_wordpress/blob/d2bfec53c3144f02f20e80fa75c118db5c635786/php/sync/class-sync-queue.php#L325 ?

php/relate/class-relationship.php
```diff
@@ -60,17 +69,20 @@ public function __construct( $post_id ) {
 	 * @return array|null
 	 */
 	public function get_data() {
-		$cache_data = wp_cache_get( $this->post_id, Sync::META_KEYS['cloudinary'] );
+		$key = $this->post_id . $this->context;
```
yuval-cloudinary, line 72: I would consider adding a separator like a dash or something
maybe adding sanitize_key() as well

yuval-cloudinary (changes requested):

php/class-utils.php
```diff
@@ -417,6 +428,24 @@ public static function upgrade_3_0_1() {
 		update_option( Sync::META_KEYS['db_version'], get_plugin_instance()->version );
 	}
 
+	/**
+	 * Upgrade DB from v3.0.1 to v3.1.9.
+	 */
+	public static function upgrade_3_1_9() {
```
yuval-cloudinary, line 434: It is interesting that the install path is a straightforward dbDelta, and the upgrade path goes through custom logic.
I would expect to go through exactly the same db scheme upgrade regardless if I uninstalled or not before upgrading

php/class-assets.php
pereirinha: That's a good question @yuval-cloudinary.

The single purpose of that query you're pointing out is to get the count of unsynced assets to populate the plugin dashboard. The way it is, it's still retrieving all the assets, regardless of the context. If we filter by context, we might get a partial overlook that doesn't reflect the overall state.

php/relate/class-relationship.php
pereirinha: I can do that, but this key is not public facing, so the benefits are questionable. What I'll do, is to ensure that the return for `get_media_context()` is sanitized.

php/class-utils.php
pereirinha: Hum, not sure if I follow.

yuval-cloudinary: sorry, after rereading the code this is a non-issue

yuval-cloudinary (approved):

PR #957, opened 2024-06-19
Fix closing the filestream and deleting the fragment

pereirinha:
Fixes the removal of unnecessary fragments.

## Approach

- Fixes the closing of the file stream;
- Fixes the function to delete the file from the file system

## QA notes

- Download an asset from Cloudinary DAM into WordPress;
- Visit the temp folder — usually `/var/tmp/` or `/tmp/`;
- Confirm that there isn't a file with the following pattern:
  - Original filename: `image.jpg`;
  - Temporary file: `image-XsWqIe.tmp` — filename suffixed with 6 random chars and the file extension tmp.

### DIFF
--- a/php/class-utils.php
+++ b/php/class-utils.php
@@ -670,7 +670,10 @@ class Utils {
 			if ( empty( $index ) ) {
 				add_action( 'shutdown', array( __CLASS__, 'purge_fragments' ) );
 			}
-			self::$file_fragments[ $index ] = $temp_file;
+			self::$file_fragments[ $index ] = array(
+				'pointer' => $pointer,
+				'file'    => $temp_file,
+			);
 			// Get the metadata of the stream.
 			$data = stream_get_meta_data( $pointer );
 			// Stream the content to the temp file.
@@ -710,8 +713,8 @@ class Utils {
 	 */
 	public static function purge_fragment( $index ) {
 		if ( isset( self::$file_fragments[ $index ] ) ) {
-			fclose( self::$file_fragments[ $index ] ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_read_fclose
-			unset( self::$file_fragments[ $index ] );
+			fclose( self::$file_fragments[ $index ]['pointer'] ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_read_fclose
+			unlink( self::$file_fragments[ $index ]['file'] );
 		}
 	}
 


php/class-utils.php
```diff
@@ -710,8 +713,8 @@ public static function purge_fragments() {
 	 */
 	public static function purge_fragment( $index ) {
```
yuval-cloudinary, line 714: Can we purge_fragment after utilizing it (I guess at Media::down_sync_asset)? That way we will not allow peak behavior, and will not be dependent on the 'shutdown' action to happen

php/class-utils.php
```diff
@@ -710,8 +713,8 @@ public static function purge_fragments() {
...
 		if ( isset( self::$file_fragments[ $index ] ) ) {
-			fclose( self::$file_fragments[ $index ] ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_read_fclose
-			unset( self::$file_fragments[ $index ] );
+			fclose( self::$file_fragments[ $index ]['pointer'] ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_read_fclose
```
yuval-cloudinary, line 716: Why are we avoiding the WP_Filesystem?

yuval-cloudinary (changes requested):

php/class-utils.php
pereirinha: @yuval-cloudinary 

We can. But as a principle, do we want to change things outside the scope of PR intent? Changing it will require additional items for QA and potentially delay the next release.

php/class-utils.php
pereirinha: That's a great question @yuval-cloudinary.

The single reason is VIP's filesystem. Please [learn more](https://docs.wpvip.com/vip-file-system/media-uploads/#h-wp-filesystem-api) here.

php/class-utils.php
yuval-cloudinary: I'm not sure about the intent. I looked at the title, and think that the current changes would not improve filestream management, as they would be closed anyway when the PHP execution is completed.
  It will improve fragment deletion, but under some edge cases, there would still be fragments left out.
  
  If this is out-of-scope, maybe we can leave a comment in the code to make sure the issue is known

php/class-utils.php
yuval-cloudinary: I guess you mean the note "most WP_Filesystem operations related to directories will not work as expected.", is it really an issue here?
If yes, does it hurt us by not being able to serve some customers, and compromise security for others?
Maybe we can detect if VIP is used, and decide if accordingly?

yuval-cloudinary (commented):
I left comments on the open issues. We might want to handle them on this PR or on a new JIRA, while keeping some comments in the codes

PR #958, opened 2024-06-19
Update wizard

pereirinha:
Updates the plugin wizard

## Approach

- Updates the print screen from the API keys page
- Updates the instructions on how to set the API keys
- Updates the build process

## QA notes

- Launch the wizard from the already existing installation;
- On a new install, you should still be guided to the wizard process


yuval-cloudinary (approved):
Looks good.

  1 comment is that screenshots are better handled by PNG than JPG

pereirinha:
@yuval-cloudinary 

I agree with the PNG vs JPG assessment for the print screen. I've made that change. While doing it, I found that some images were not part of the build process, and some trash had been left behind.

We are now taking care of it.

Mind to circle back?

Thank you.

yuval-cloudinary (approved):

PR #967, opened 2024-08-21
Prevent the duplication of Cloudinary's analytics query arg

pereirinha:
## Approach

- By prepending the base URL with Cloudinary's query argument in the search alias array, we'll ensure that it will try first to replace the URLs where this argument is part of the existing URL.

### DIFF
--- a/php/class-delivery.php
+++ b/php/class-delivery.php
@@ -1016,11 +1016,13 @@ class Delivery implements Setup {
 				continue;
 			}
 			if ( ! empty( $relation['slashed'] ) && $relation['slashed'] ) {
-				$aliases[ $base . '?' ] = addcslashes( $cloudinary_url . '&', '/' );
-				$aliases[ $base ]       = addcslashes( $cloudinary_url, '/' );
+				$aliases[ $base . '?_i=AA' ] = addcslashes( $cloudinary_url, '/' );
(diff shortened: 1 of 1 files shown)


php/class-delivery.php
```diff
@@ -1016,11 +1016,13 @@ public function convert_tags( $content, $context = 'view' ) {
 				continue;
 			}
 			if ( ! empty( $relation['slashed'] ) && $relation['slashed'] ) {
-				$aliases[ $base . '?' ] = addcslashes( $cloudinary_url . '&', '/' );
-				$aliases[ $base ]       = addcslashes( $cloudinary_url, '/' );
+				$aliases[ $base . '?_i=AA' ] = addcslashes( $cloudinary_url, '/' );
```
tharsheblows, line 1019: **❓question**
is there any chance there would be more query variables? Adding more also causes the sitemap to fail.

tharsheblows (approved):
There's a question about whether or not there will be more query variables in the query string. It makes the sitemap file if there are; I added mine to the beginning of the query string.

php/class-delivery.php
pereirinha: @tharsheblows 

Great finding.

My testing will depend on whether the `_i` argument is the first. If so, this will still work fine; otherwise, it will fail.

These should be edge cases; we might have better ways to work around them.

Let's follow up internally.

PR #969, opened 2024-08-21
Clean up errored assets

pereirinha:
## Approach

- Adds means to clean up the sync error flag, allowing the assets to be retried to sync. This can be achieved by one of these:
  - Via the dashboard;
  - Via the Media Library;
  - Via the CLI commands


php/sync/class-upload-sync.php
```diff
@@ -91,7 +90,8 @@ private function register_hooks() {
 			'bulk_actions-upload',
 			function ( $actions ) {
 				$cloudinary_actions = array(
-					'cloudinary-push' => __( 'Sync with Cloudinary', 'cloudinary' ),
+					'cloudinary-push'     => __( 'Sync with Cloudinary', 'cloudinary' ),
+					'cloudinary-clean-up' => __( 'Clean Cloudinary error', 'cloudinary' ),
```
tharsheblows, line 94: **🥜 nitpick**
This is so small but I think "error" should be capitalised to keep it consistent?

php/sync/class-upload-sync.php
```diff
@@ -136,18 +136,35 @@ public function add_inline_action( $actions, $post ) {
...
+				add_query_arg(
+					array(
+						'action'   => 'cloudinary-clean-up',
+						'media[]'  => $post->ID,
+						'_wpnonce' => wp_create_nonce( 'bulk-media' ),
+					),
+					admin_url( 'upload.php' )
+				),
+				esc_attr__( 'Clean Cloudinary error', 'cloudinary' ),
+				esc_html__( 'Clean Cloudinary error', 'cloudinary' )
```
tharsheblows, line 164: **🥜 nitpick**
Same here, maybe capitalise Error to keep it consistent

tharsheblows (approved):
There are two very small comments regarding capitalisation. If you think they don't matter then you can ignore!

I've tested all three scenarios also and they are working.

php/sync/class-upload-sync.php
pereirinha:
```suggestion
					'cloudinary-clean-up' => __( 'Clean Cloudinary Error', 'cloudinary' ),
```

php/sync/class-upload-sync.php
pereirinha:
```suggestion
				esc_html__( 'Clean Cloudinary Error', 'cloudinary' )
```

pereirinha:
> There are two very small comments regarding capitalisation. If you think they don't matter then you can ignore!
> 
> I've tested all three scenarios also and they are working.

Thank you @tharsheblows. I took care of these.

Issue #933, opened 2024-02-05
This is failing and returning notice error


shay-te:
Hello this function seems to be broken in our project. 

https://github.com/cloudinary/cloudinary_wordpress/blob/edd640e708893f77002d92c9c83ac8fd82163772/php/class-utils.php#L518

it returns this notice in the start of JSON
```
<br />
<b>Notice</b>:  Uninitialized string offset: 0 in <b>/var/www/html/wp-content/plugins/cloudinary-image-management-and-manipulation-in-the-cloud-cdn/php/class-utils.php</b> on line <b>518</b><br />
```

We fixed it like this:

```
         */
        public static function looks_like_json( $thing ) {
//              return ! empty( $thing ) && is_string( $thing ) && in_array( ltrim( $thing )[0], array( '{', '[' ), true );
              if (is_string($thing)) {
                  $trimmedThing = ltrim($thing);
                  return !empty($thing) && strlen($trimmedThing) > 0 && in_array($trimmedThing[0], ['{', '['], true);
              } else {
                  return false; // or handle the case when $thing is not a string
              }
        }
```

Vdeub-cloudinary:
Hi @shay-te, 

Would you mind sharing more context on the issue? Is there a way to reproduce it on our side? 

Could you also share a system report as documented [here](https://support.cloudinary.com/hc/en-us/articles/360022260599-How-to-generate-a-system-report-on-the-WordPress-plugin)?  If you don't want to share it here, please open a ticket to support@cloudinary.com.

Thanks!

Vdeub-cloudinary:
Hi @shay-te, is still relevant? 
Thanks!

shay-te:
No

Vdeub-cloudinary:
Thanks @shay-te !

PR #970, opened 2024-08-22
fix: prevent php warning in looks_like_json function

epigeyre:
If `$thing` only contains a `space` it will generate a php warning when doing `ltrim( $thing )[0]`. This fixes the issue while still being compatible with php 5.4+.

Fixes issue #933

### DIFF
--- a/php/class-utils.php
+++ b/php/class-utils.php
@@ -543,7 +543,21 @@ class Utils {
 	 * @return bool
 	 */
 	public static function looks_like_json( $thing ) {
-		return is_string( $thing ) && ! empty( ltrim( $thing ) ) && in_array( $thing[0], array( '{', '[' ), true );
+		if ( ! is_string( $thing ) ) {
+			return false;
+		}
+
+		$thing = trim( $thing );
+	
+		if ( empty( $thing ) ) {
+			return false;
+		}
+	
+		if ( ! in_array( $thing[0], array( '{', '[' ), true ) ) {
+			return false;
+		}
+
+		return true;
 	}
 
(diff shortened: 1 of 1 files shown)


php/class-utils.php
```diff
@@ -543,7 +543,7 @@ public static function pathinfo( $path, $flags = 15 ) {
 	 * @return bool
 	 */
 	public static function looks_like_json( $thing ) {
-		return ! empty( $thing ) && is_string( $thing ) && in_array( ltrim( $thing )[0], array( '{', '[' ), true );
+		return is_string( $thing ) && ! empty( ltrim( $thing ) ) && in_array( $thing[0], array( '{', '[' ), true );
```
pereirinha, line 546: **👏 praise:**

@epigeyre thank you for the PR. However, it will fail to test if the JSON will look like this ` {}`.

Also, I'm realizing that the cognitive load for this method is a bit too much, making it hard to read.

I'm adding a suggestion here. Would you mind to give it a try?

```suggestion
		if ( ! is_string( $thing ) ) {
			return false;
		}

		$thing = trim( $thing );
	
		if ( empty( $thing ) ) {
			return false;
		}
	
		if ( ! in_array( $thing[0], array( '{', '[' ), true ) ) {
			return false;
		}

		return true;
```

Note that in your approach, you're not trimming the string before the last check, and that's why it will return `false` on the test I've made.

I'll be happy to add it to the upcoming release.

epigeyre: @pereirinha You are right this is much better. Thanks !

PR #979, opened 2024-09-11
Wizard improvements

pereirinha:
- Prevents accessing the connection string from the wizard
- Fixes an issue cleaning up this information from the local storage
- Fixes the infinite loop while saving on the last step of the wizard
- Fixes the locked state for the toggles on the wizard


tharsheblows (approved):
@pereirinha This all looks good! I've tested. There's a merge conflict so I haven't merged but it's good to do so once that's sorted.

PR #981, opened 2024-09-17
Discard faulty params

pereirinha:
- Prevents fatal error when there's corrupted data

### DIFF
--- a/php/traits/trait-params.php
+++ b/php/traits/trait-params.php
@@ -53,10 +53,15 @@ trait Params_Trait {
 			return $new;
 		}
 
-		if ( '' === $key ) {
-			$new[] = $value;
+		if ( is_array( $new ) ) {
+			if ( '' === $key ) {
+				$new[] = $value;
+			} else {
+				$new[ $key ] = $value;
+			}
 		} else {
-			$new[ $key ] = $value;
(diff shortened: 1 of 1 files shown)


php/traits/trait-params.php
```diff
@@ -53,10 +53,15 @@ protected function set_param_array( $parts, $param, $value ) {
...
+		if ( is_array( $new ) ) {
+			if ( '' === $key ) {
+				$new[] = $value;
+			} else {
+				$new[ $key ] = $value;
+			}
 		} else {
-			$new[ $key ] = $value;
+			// There's likely a problem here, as we're setting an array key on a non-array.
+			$new = array();
```
tharsheblows, line 64: **❓question**
I'm approving this but want to check that once `$new` is an array, you don't want to set the key/value on it?

tharsheblows (approved):
@pereirinha I have one question but it's good to merge if nothing needs to be done.

php/traits/trait-params.php
pereirinha: That's correct. If `$param` and hence `$new` is a string, then that three is broken, and we are better off ignoring it. Upon a new save, it should be rebuilt accordingly. It's as if a new setting was introduced.

PR #982, opened 2024-09-18
Improves the workflow for updating the connection string

pereirinha:
- Prevents updating the connection string with an invalid value
- Allows to cancel the update process


php/class-media.php
```diff
@@ -1549,7 +1549,8 @@ public function prepare_size( $attachment_id, $size ) {
 			}
 		} elseif ( is_string( $size ) || ( is_array( $size ) && 3 === count( $size ) ) ) {
 			$intermediate = image_get_intermediate_size( $attachment_id, $size );
-			if ( is_array( $intermediate ) ) {
+			// PDF's do not have intermediate URL.
+			if ( is_array( $intermediate ) && ! empty( $intermediate['url'] ) ) {
```
tharsheblows, line 1553: **❓question**
You didn't mention this so I'm double checking it's supposed to be in here. Not a blocker!

tharsheblows (approved):
@pereirinha I've tested this with the wizard and it works for me. I don't know about this file `php/media.php` so want to check it's part of this. Otherwise approved.

PR #987, opened 2024-10-15
When on read only mode, do not assume that the checkbox is checked

pereirinha:


tharsheblows (approved):
@pereirinha This looks good and works well for me!

PR #988, opened 2024-10-15
Wrap `rest_url`, `site_url` and `home_url` so that we can filter these securely

pereirinha:


tharsheblows (commented):
@pereirinha Should every instance of `home_url()`, `site_url()` and `rest_url()` be filtered in the plugin? I'm asking because I'll check that too if you'd like.

This looks good to me though, I like the approach! It keeps it to the plugin nicely.

PR #997, opened 2024-11-12
Math does not allow you to divide by zero

pereirinha:
Prevent fatal error when there are no assets in the media library


tharsheblows (approved):
Looks good! I checked to make sure everything is a number too and it is. :)

PR #999, opened 2024-11-22
Fix: remove custom babel config causing issues in WP 6.6+

tharsheblows:
This removes `.babelrc` as it was causing issues and I think is no longer needed.

The Gallery block no longer has a dependency on `wp-polyfills` although testing is needed to see if that should be conditionally added in WP < 6.6.


pereirinha (approved):
@tharsheblows 

This looks good. 
I've reviewed the changes and tested WP 6.0

PR #1003, opened 2024-12-06
Add support for `source` `screset`

pereirinha:


php/class-delivery.php
```diff
@@ -1365,14 +1365,14 @@ public function parse_element( $element ) {
...
-		$third_party_change  = Utils::maybe_get_third_party_changes( $attributes );
+		$tag_element['type'] = in_array( $tag_element['tag'], array( 'img', 'source' ), true ) ? 'image' : $tag_element['tag'];
+		$third_party_change  = Utils::maybe_get_third_party_changes( $attributes, $tag_element['tag'] );
 		if ( ! empty( $third_party_change ) ) {
 			Utils::log( $third_party_change, 'third-party-loading' );
 
 			return null;
 		}
-		$raw_url                 = $attributes['src'];
+		$raw_url                 = 'source' === $tag_element['tag'] ? $attributes['srcset'] : $attributes['src'];
```
tharsheblows, line 1375: **❗️issue**
I'm getting warnings here which come from `wp-includes/media-template.php` [eg here](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/media-template.php#L693) and the video and audio `<source>` tags.

tharsheblows (changes requested):
I'm getting some warnings in the editor which result from the WP media template's <audio> and <video> <source> tags.

php/class-delivery.php
pereirinha: @tharsheblows 

I did push an update. 

Can you please have a look? :)

PR #1006, opened 2024-12-20
Use the `latest`  video player

pereirinha:
Update the video player


tharsheblows (commented):
There are styles and scripts which use this as a version string which means that it will update without changing those and it could pick up cached versions inconsistently. Should those be updated also or maybe that's a reason to hold it to a given version?

tharsheblows (approved):
Thank you!

PR #1019, opened 2025-02-24
Experimental support for GLB models.

dero:
Must be enabled by hooking into a filter in code.

```php
add_filter( 'cloudinary_allow_glb_upload', '__return_true' );
```

### DIFF
--- a/php/media/class-gallery.php
+++ b/php/media/class-gallery.php
@@ -30,6 +30,15 @@ class Gallery extends Settings_Component {
 	const GALLERY_LIBRARY_HANDLE = 'cld-gallery';
 
 	/**
+	 * This is a "false" mime type for GLB files.
+	 *
+	 * Canonically, the correct mime type for GLB files is `model/gltf-binary`,
+	 * however we need to fake an `image/` mime type to make the gallery
+	 * picker to list GLB files.
+	 */
+	const GALLERY_GLB_MIME_TYPE = 'image/glb';
+
+	/**
 	 * Holds the settings slug.
 	 *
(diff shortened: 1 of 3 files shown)


php/media/class-gallery.php
```diff
@@ -568,6 +583,105 @@ function_exists( 'has_block' ) && has_block( 'cloudinary/gallery' )
...
+	 * @return string
+	 */
+	public function admin_replace_glb_url( $url, $attachment_id ) {
+		if ( ! is_admin() ) {
+			return $url;
+		}
+
+		if ( self::GALLERY_GLB_MIME_TYPE === get_post_mime_type( $attachment_id ) ) {
+			$url = str_replace( '.glb', '.png', $url );
+		}
```
tharsheblows, line 663: **❓question**
The 3D model isn't viewable at all in the admin area? Should we document this somewhere, maybe when adding in the hook for enabling this? It should show on preview though?

tharsheblows (commented):
This seems to work only on gallery blocks for me, not on any other block type, eg `image`. Is that correct?

php/media/class-gallery.php
dero: @tharsheblows It is documented on the hook, [see here](https://github.com/cloudinary/cloudinary_wordpress/pull/1019/files#diff-55f3118701b75994c7ff51958f717de85e006698c9044621a84a454073373e95R699-R700).

Do you think we should make the comment more explicit?

tharsheblows (approved):

PR #1020, opened 2025-03-03
Ensure sources are an array in srcset hook.

dero:


php/class-media.php
```diff
@@ -1874,6 +1874,12 @@ public function image_srcset( $sources, $size_array, $image_src, $image_meta, $a
 		if ( ! $cloudinary_id ) {
 			return $sources; // Return WordPress default sources.
 		}
+
+		// Handle unexpected sources variable type.
+		if ( ! is_array( $sources ) ) {
```
tharsheblows, line 1879: **❓question**
Does `$sources` ever come into this as eg an empty string and then go through and get set via the metadata? [Just below this there is a check on the a lot of the metadata to set it](https://github.com/cloudinary/cloudinary_wordpress/blob/master/php/class-media.php#L1888-L1891) where `$sources` is set.

The check should happen [after that](https://github.com/cloudinary/cloudinary_wordpress/blob/master/php/class-media.php#L1927) I think, from that point on it's assuming `$sources` is an array

tharsheblows (commented):
One question about the position of the array check.

php/class-media.php
dero: @tharsheblows Technically, `$sources` should ever be anything else than an array:

* https://developer.wordpress.org/reference/hooks/wp_calculate_image_srcset/

The hook clearly documents `$sources` should always be an array. The intent here was just safeguard against emitting a PHP warning in case something unexpected happens and `$sources` comes in as a non-array (as it does on one of the customers' site).

Having said that, I do agree with you that it's smarter to move the check after we've inspected post meta as it's entirely possible we'll end up resetting `$sources` there. Making the change.

dero: Moved in [3fe0646](https://github.com/cloudinary/cloudinary_wordpress/pull/1020/commits/3fe06467fae1f6f0070eb326884a24ee4435ca61).

tharsheblows (approved):

Issue #1022, opened 2025-03-20
Add ability to remove dns-prefetch or preconnect


tdatu:
## Feature description

Adding ability to remove dns-prefetch or preconnect via filter hooks.

aatanasov-cloudinary:
Hi @tdatu , thanks for your contribution. The filter will be available in the next release.

aatanasov-cloudinary:
The filter is now available and documented in [here](https://cloudinary.github.io/cloudinary_wordpress/cloudinary_dns_prefetch_types.html).

PR #1023, opened 2025-03-20
add ability to remove dns-prefetch or preconnect link tags

tdatu:
Issue: #1022 

## Approach

Implemented apply_filters in setup_hooks

## QA notes

Simple hooks and implemented in test environment.

### DIFF
--- a/php/class-delivery.php
+++ b/php/class-delivery.php
@@ -765,6 +765,8 @@ class Delivery implements Setup {
 	 */
 	public function dns_prefetch( $urls, $relation_type ) {
 
+		$relation_type = apply_filters('cld_dns_prefetch', $relation_type);
+
 		if ( 'dns-prefetch' === $relation_type || 'preconnect' === $relation_type ) {
 			$urls[] = $this->media->base_url;
 		}
--- a/php/class-delivery.php
+++ b/php/class-delivery.php
@@ -765,11 +765,17 @@ class Delivery implements Setup {
 	 */
 	public function dns_prefetch( $urls, $relation_type ) {
 
-		$relation_type = apply_filters('cld_dns_prefetch', $relation_type);
+		/**
+		 * Filter to provide option to omit prefetch.
+		 *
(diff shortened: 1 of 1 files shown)


atdcloud:
@tharsheblows, please check.

php/class-delivery.php
```diff
@@ -765,6 +765,8 @@ protected function setup_hooks() {
 	 */
 	public function dns_prefetch( $urls, $relation_type ) {
 
+		$relation_type = apply_filters('cld_dns_prefetch', $relation_type);
```
tharsheblows, line 768: **📝 note**
This will work but I think the way to do it is something like below. It keeps it a bit more understandable when looking at the code.

```
... 
$dns_prefetch_types = apply_filters( 'cld_dns_prefetch_types', array ( 'dns-prefetch', 'preconnect' ) );

if ( in_array( $relation_type, $dns_prefetch_types, true ) { 
...
```

tdatu: hmmm...I think the IF clause that evaluates the $relation_type is already a dead giveaway.

dero: I think we should do a couple of things here:

1. Avoid using the words `prefetch_types` in filter and variable names, as it's a misnomer. Let's use `resource_hints`.
2. I support using `in_array` to explicitly search through an array instead of mutating `$relation_type`. In fact, mutating it has no purpose, as the `$relation_type` sets the context for the entire method and should, in my opinion, be treated as immutable.

```php
...
$resource_hints = apply_filters( 'cld_resource_hints', array ( 'dns-prefetch', 'preconnect' ) );

if ( in_array( $relation_type, $resource_hints, true ) { 
...
```

aatanasov-cloudinary (approved):
@tdatu, Thank you for your contribution. Looks good, I'll adjust the Filter information in another PR. That way, the new filter will appear [here](https://cloudinary.github.io/cloudinary_wordpress/cloudinary_init_delivery.html).

PR #1026, opened 2025-03-26
Fix DivisionByZeroError in Responsive_Breakpoints

aatanasov-cloudinary:
Fixing `DivisionByZeroError` occurring in the `Responsive_Breakpoints` class.

## Approach

 - If an attachment doesn't have `width` or `height`, the initial image tag will be rendered on the front-end without any srcset information.
 - Add a debug log if there is an issue with `width` and `height`.
 - Remove a debug log during syncing attachments.

## QA notes

 - Upload an image.
 - Set its height or width to `0` (during the tests a database update was followed)
 - Insert the image to a page.
 - Save and preview the page.
 - The image should appear without errors.

### DIFF
--- a/php/delivery/class-responsive-breakpoints.php
+++ b/php/delivery/class-responsive-breakpoints.php
@@ -127,12 +127,19 @@ class Responsive_Breakpoints extends Delivery_Feature {
 	 * @return array
 	 */
 	protected function apply_breakpoints( $tag_element ) {
+		$settings  = $this->settings->get_value( 'media_display' );
+		$max       = $settings['max_width'];
+		$min       = $settings['min_width'];
+		$width     = (int) $tag_element['width'];
+		$height    = (int) $tag_element['height'];
+		$debug_key = 'responsive_breakpoints_' . $tag_element['id'];
 
-		$settings            = $this->settings->get_value( 'media_display' );
-		$max                 = $settings['max_width'];
-		$min                 = $settings['min_width'];
-		$width               = $tag_element['width'];
-		$height              = $tag_element['height'];
+		if ( ! $width || ! $height ) {
+			// Translators: %s is the ID of the image missing width or height.
+			Utils::log( sprintf( __( 'Missing width or height for image: %s', 'cloudinary' ), $tag_element['id'] ), $debug_key );
+			return $tag_element;
+		}
+
 		$size_tag            = '-' . $width . 'x' . $height . '.';
 		$step                = $settings['pixel_step'];
 		$ratio               = $width / $height;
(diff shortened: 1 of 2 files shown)


php/delivery/class-responsive-breakpoints.php
```diff
@@ -159,6 +166,8 @@ protected function apply_breakpoints( $tag_element ) {
 			$tag_element['atts']['sizes']  = '(max-width: ' . $width . 'px) 100vw, ' . $width . 'px';
 		}
 
+		Utils::remove_log( $debug_key );
```
dero, line 169: 💡 **Thought:** I can imagine `apply_breakpoints` could be called from multiple places, and only one of those places would provide faulty data. In such a case, we would probably want the log message to be retained, not removed by subsequent correct invocations.

The debug log is only ever consulted for clues and is manually reviewed and interpreted. I think we can afford to keep these messages in even after we _think_ the issue has been resolved.

php/delivery/class-responsive-breakpoints.php
```diff
@@ -127,12 +127,19 @@ public function add_features( $tag_element ) {
...
 	protected function apply_breakpoints( $tag_element ) {
+		$settings  = $this->settings->get_value( 'media_display' );
+		$max       = $settings['max_width'];
+		$min       = $settings['min_width'];
+		$width     = (int) $tag_element['width'];
+		$height    = (int) $tag_element['height'];
+		$debug_key = 'responsive_breakpoints_' . $tag_element['id'];
+
+		if ( ! $width || ! $height ) {
+			// Translators: %s is the ID of the image missing width or height.
```
dero, line 138: 💬 **suggestion:** ‏I think we shouldn't translate this string. It will only ever be used by English-speaking people, and we don't want to be logging messages in other languages.

php/delivery/class-responsive-breakpoints.php
```diff
@@ -127,12 +127,19 @@ public function add_features( $tag_element ) {
 	 * @return array
 	 */
 	protected function apply_breakpoints( $tag_element ) {
+		$settings  = $this->settings->get_value( 'media_display' );
+		$max       = $settings['max_width'];
+		$min       = $settings['min_width'];
+		$width     = (int) $tag_element['width'];
+		$height    = (int) $tag_element['height'];
+		$debug_key = 'responsive_breakpoints_' . $tag_element['id'];
```
dero, line 135: 💬 **suggestion:** ‏Since we're hardening the contract here, we might as well make sure `$tag_element['id']` is also set before using it.

php/delivery/class-responsive-breakpoints.php
aatanasov-cloudinary: @dero, thanks, I agree; this is also something that I considered as an approach. I'm going to remove the `remove_log` logic and its related code.

tharsheblows (approved):
Looks great, thank you!

PR #1031, opened 2025-04-17
Ensure single and double quotes are properly used when buiding img tags.

aatanasov-cloudinary:
Some Page Builders use HTML markup inside JSON, which causes issues with single/double quotes when building our img tag.

## Approach

- Depending on the provided HTML tag, we use single or double quotes for our HTML element.

## QA notes

- Force the HTML tag to include single quotes (you can mock this locally).
- Do the same for double quotes to ensure it works.

### DIFF
--- a/php/class-delivery.php
+++ b/php/class-delivery.php
@@ -1376,6 +1376,12 @@ class Delivery implements Setup {
 		// Setup new tag.
 		$replace = HTML::build_tag( $tag_element['tag'], $tag_element['atts'] );
 
+		// If the original tag used single quotes, we need to use them in the new tag.
+		$single_quotes = (bool) preg_match( '/=\s*\'/', $tag_element['original'] );
+		if ( $single_quotes ) {
+			$replace = str_replace( '"', "'", $replace );
+		}
+
 		/**
 		 * Filter the new built tag element.
 		 *


php/class-delivery.php
```diff
@@ -1376,6 +1376,12 @@ public function rebuild_tag( $tag_element ) {
 		// Setup new tag.
 		$replace = HTML::build_tag( $tag_element['tag'], $tag_element['atts'] );
 
+		// If the original tag used single quotes, we need to use them in the new tag.
+		$single_quotes = (bool) preg_match( '/=\s*\'/', $tag_element['original'] );
+		if ( $single_quotes ) {
+			$replace = str_replace( '"', "'", $replace );
+		}
```
tharsheblows, line 1383: **❓question**
Will this break if someone adds a single quote to eg the alt attribute? Like 
`They're the best birds`

I'm unsure on how to test that.

tharsheblows (approved):
@aleksandar-cloudinary One question but approving!

php/class-delivery.php
aleksandar-cloudinary: @tharsheblows I think you meant @aatanasov-cloudinary 😄

aatanasov-cloudinary: Hi @tharsheblows , thanks, I've tested this, and WordPress uses HTML Entities for text attributes like `alt`, and they are rendered correctly. Here is an example with single quotes and alt text:

alt-text

PR #1032, opened 2025-04-24
Convert maybe_unset_attributes to filter

aatanasov-cloudinary:
The way our plugin uses the `rest_request_before_callbacks` hook causes issues with Rest API requests validation.

## Approach

- We now use a filter instead of an action.
- We ensure the `$response` is passed when using the `rest_request_before_callbacks` filter.

## QA notes

 1. Add the below test filter to a plugin.
 2. Go to `/wp-json/wp/v2/media/`.
 3. The WP Error should appear. Previously, it was skipped. 

Test filter:
```
add_filter( 'rest_request_before_callbacks', function( $response, $handler, $request) {
	return new \WP_Error( 'error', __( 'Force an error to test the filter', "my_textdomain" ) );
}, 7, 3 );
```


tharsheblows (approved):
Thank you for fixing my mistake!

PR #1043, opened 2025-05-20
Fix _load_textdomain_just_in_time notice

### LINKED ISSUE #1038: _load_textdomain_just_in_time called too early – triggers notice in WP 6.8
**Title:**  
`_load_textdomain_just_in_time` called too early – triggers notice in WP 6.8

**Describe the bug**  
After updating WordPress to version 6.8, the following notice appears when using the Cloudinary plugin:

Notice: Function _load_textdomain_just_in_time was called incorrectly.
Translation loading for the cloudinary domain was triggered too early.
This is usually an indicator for some code in the plugin or theme running too early.
Translations should be loaded at the init action or later.
Please see Debugging in WordPress for more information.
(This message was added in version 6.7.0.)
in /application/www/wp/wp-includes/functions.php on line 6121

**Steps to reproduce**
1. Update to WordPress 6.8.
2. Enable `WP_DEBUG`.
3. Activate and use the Cloudinary plugin.
4. The notice appears immediately on load.

**Expected behavior**  
Translations should be loaded later in the WordPress lifecycle (e.g., during the `init` action) to comply with changes introduced in WordPress 6.7+.

**Environment:**
- WordPress version: 6.8  
- Cloudinary plugin version: 3.2.7
- PHP version: 8.2

**Additional context**  
The issue is caused by calling `load_plugin_textdomain()` too early. Since WP 6.7, loading translation domains before the `init` hook results in a warning.

aihowes:
Adds set_cloudinary_filters to the init hook.

Fixes #1038

## Approach

- The i18n function `__()` was being called in the constructor of the class which is considered to early
- I have moved the relevant code to a separate function and added the function as an action to take place on init.

## QA notes

- Detail the steps needed to verify the PR.

### DIFF
--- a/php/class-media.php
+++ b/php/class-media.php
@@ -158,6 +158,19 @@ class Media extends Settings_Component implements Setup {
 	public function __construct( Plugin $plugin ) {
 		$this->plugin = $plugin;
 
+		// Add set cloudinary filters to init so i18n methods are usable.
+		add_action( 'init' , array( $this, 'set_cloudinary_filters' ) );
+
+		// Add upgrade hook, since setup methods are called after the connect upgrade has run.
+		add_action( 'cloudinary_version_upgrade', array( $this, 'upgrade_media_settings' ) );
+	}
+
+	/**
+	 * Media constructor.
+	 *
+	 * @param Plugin $plugin The global plugin instance.
+	 */
+	public function set_cloudinary_filters() {
 		/**
 		 * Filter the Cloudinary Media Library filters.
 		 *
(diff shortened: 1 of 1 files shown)


aatanasov-cloudinary:
@aihowes, thank you for your contribution. 

We have a fix for this in [develop](https://github.com/cloudinary/cloudinary_wordpress/commits/develop/) with an addition `_load_textdomain_just_in_time` optimization, which will be released in the next version. We're planning to release next week.

I'll close this PR, but we appreciate the time you spent on this issue.

PR #1072, opened 2025-07-15
Fix/plugin assets sync

aatanasov-cloudinary:
Fixes the following:
 * Syncing plugin (parent) assets when they have new versions.
 * Adding a version to the plugin assets to ensure no old cached assets are delivered.
 * Ensuring the icons from the Cloudinary font render correctly.

## Approach

 - For the Cloudinary font, the `css-unicode-loader` package was installed to ensure such code isn't affected: `content: "\e900";`
 - For the parent syncing issue, the following has been done:
    - Set the version to the asset URL using the `get_cloudinary_version` method.
    - Added a new method called `sync_parent` to take care of syncing parent assets.

## QA notes

 - Install a test plugin like this one:  [test-image-updated.zip](https://github.com/user-attachments/files/21233381/test-image-updated.zip). Please rename the plugin folder and the plugin name.
 - Go to Cloudinary General Settings and ensure the plugin is enabled for syncing: test-plugin-sync (note the disable asset settings don't work, this issue will be addressed in another PR).
 - Go to the admin page of the test plugin so Cloudinary syncs the plugin assets.
 - Wait for a minute, and the images should be delivered from Cloudinary. Something like that: `https://res.cloudinary.com/example-account/images/w_500/v248/cloudinary.local/wp-content/plugins/test-image-updated/example1-test-image/example1-test-image.png?_i=AA` -- -`v248` is the version of the plugin: `2.48`
 - Change one of the test images in the test plugin, but the file name must be the same.
 - Increase the test plugin version.
 - Go again to the admin page of the test plugin and wait for a few minutes.
 - After a few minutes, the new version of the image should be available, and the version in the image URL to be changed.
 - Ensure there are no debug errors.

### DIFF
--- a/php/class-assets.php
+++ b/php/class-assets.php
@@ -156,7 +156,7 @@ class Assets extends Settings_Component {
 		add_filter( 'cloudinary_asset_state', array( $this, 'filter_asset_state' ), 10, 2 );
 		add_filter( 'cloudinary_set_usable_asset', array( $this, 'check_usable_asset' ) );
 		// Actions.
-		add_action( 'cloudinary_init_settings', array( $this, 'setup' ) );
+		add_action( 'cloudinary_ready', array( $this, 'setup' ) );
 		add_action( 'cloudinary_thread_queue_details_query', array( $this, 'connect_post_type' ) );
 		add_action( 'cloudinary_build_queue_query', array( $this, 'connect_post_type' ) );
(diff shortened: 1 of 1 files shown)


gabrielcld2 (approved):
Changes look good. I was able to follow the testing steps locally and can confirm the issue is fixed ✅

PR #1075, opened 2025-07-18
Improve unknown URLs search

aatanasov-cloudinary:
The plugin has a feature where it detects if there are unknown URLs and tries to sync them if they exist in the database ([code reference](https://github.com/cloudinary/cloudinary_wordpress/blob/master/php/class-delivery.php#L1950-L1952)).  When there are unknown URLs there is [a postmeta query](https://github.com/cloudinary/cloudinary_wordpress/blob/master/php/class-delivery.php#L896).

The main issue was that attachment sizes are considered unknown, which is incorrect. Because of that, the postmeta query was executed on every page request. For large websites, this can be an issue.

## Approach

- Added a method to find truly unknown files.

## QA notes

- Install Query Monitor.
- Go to a page with attachments.
- This query shouldn't appear: postmeta-query
 - Go to `wp-content/uploads/2025/07/` and add an image file (e.g. `testing-image-sync.jpg`). This should happen outside of WordPress.
 - After that, using the code editor, insert this image tag: `` (set the correct src based on your instance) to a page.
 - Save the page.
 - Go to the page on the front-end to confirm the test image isn't delivered from Cloudinary.
 - Using PHP script, insert [the attachment](https://developer.wordpress.org/reference/functions/wp_insert_attachment/) (outside of WP Media Library). This is a script generated with Copilot: [insert-attachment.php.zip](https://github.com/user-attachments/files/21312400/insert-attachment.php.zip) -- open the admin page only once.
 - Go to the page again, and the `Cloudinary\Delivery->find_attachment_size_urls()` query should appear in Query Monitor.
 - Wait for a minute, and refresh the page. 
 - Finally, the previously unknown image should be delivered from Cloudinary, and the postmeta query shouldn't appear.

### DIFF
--- a/php/class-delivery.php
+++ b/php/class-delivery.php
@@ -883,6 +883,7 @@ class Delivery implements Setup {
 		$dirs    = wp_get_upload_dir();
 		$baseurl = Utils::clean_url( $dirs['baseurl'] );
 		$search  = array();
+
 		foreach ( $this->unknown as $url ) {
 			$url      = ltrim( str_replace( $baseurl, '', $url ), '/' );
 			$search[] = $url;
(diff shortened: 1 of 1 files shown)


gabrielcld2 (approved):
The code looks good and following the reproduction steps worked 👍

PR #1076, opened 2025-07-25
Fix audio source tags

aatanasov-cloudinary:
Fixes audio source tags.

## Approach

- Ensuring audio source tags aren't converted.

## QA notes

 - Go to a post and insert an audio file.
 - The audio file should work correctly on the front-end.
 - Install the classic editor.
 - Go to a new page and insert an audio file.
 - The audio file should work correctly on the front-end.
 - There shouldn't be any errors.


gabrielcld2 (approved):
Works as expected ✅

PR #1080, opened 2025-08-07
Fix webm audio files

aatanasov-cloudinary:
Fixes webm files that have only audio.

## Approach

- If videos are delivered, we check the mime type of the video. 
- If it is `audio/webm`, we don't require width and height.

## QA notes

- Upload a webm file that has only audio.
- Ensure the file is synced.


gabrielcld2 (approved):
👍 working as expected

PR #1085, opened 2025-08-22
Avoid Sync Types Fatal Errors

gabrielcld2:
This ensures that even if the sync types aren't loaded, there's no fatal errors due to sync_types being null.

## Approach

Main reason: Current fatal errors occurs because the `sync_types` property is set to null (instead of an array). By default, properties in PHP are initialized as null unless told otherwise. By initializing it to an empty array, we can guarantee these fatal errors won't occur.

More detailed investigation of the reported issue:
- This `sync_types` property is assigned only in one place, which is the `setup_sync_types` method, which comes originally from this `init` filter with a priority of **200** https://github.com/cloudinary/cloudinary_wordpress/blob/master/php/class-plugin.php#L278 
- The reported broken site is using another plugin called [Rank Math SEO](https://wordpress.org/plugins/seo-by-rank-math/) which initialize its settings page on an `init` filter of priority **125**. This means this executes before `sync_types`
- That plugin calls `wp_get_attachment_image_src` within its settings page, which in turns calls our own cloudinary filter, resulting in the sync_types variable being called before it is initialized.
- The needs for this plugin to call this this early is simply to build a preview of the google search results. It supports a way to add a link to the featured image, which when added causes that issue.

Here's a stack trace of the issue (using XDebug)
Screenshot 2025-08-22 at 10 07 00

## QA notes

Two approaches:

1. Import the backup of the site having the issue (a bit long as the site is quite large)
2. Follow the manual steps below to reproduce it:

- Install and activate the `Rank Math SEO` plugin
- Configure it using its Wizard
  - create a free account
  - set the installer to "easy" mode and leave the other values as default
- Create a new post
  - add a post title and a featured image to it 
- On the editor, there should be a new button for the Rank Math SEO in the top right, click on it
  - Then click on "Edit Snippet"
Screenshot 2025-08-22 at 11 08 13
  - Within the post description field, input the `%post_thumbnail%` variable
  - Save the changes for the snippet and publish the post
- Without the current fix, the site should now crash
- With the fix, the site should work & the snippet preview should show the cloudinary link properly

### DIFF
--- a/php/class-sync.php
+++ b/php/class-sync.php
@@ -372,10 +372,13 @@ class Sync implements Setup, Assets {
 				$signature = array();
 			}
 
-			// Remove any old or outdated signature items. against the expected.
-			$signature                    = array_intersect_key( $signature, $this->sync_types );
 			$signatures[ $attachment_id ] = $return;
-			$return                       = wp_parse_args( $signature, $this->sync_types );
+
+			// Remove any old or outdated signature items. against the expected.
+			if ( ! empty( $this->sync_types ) ) {
+				$signature = array_intersect_key( $signature, $this->sync_types );
+				$return    = wp_parse_args( $signature, $this->sync_types );
+			}
 		}
 
 		/**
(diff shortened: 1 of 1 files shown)


aatanasov-cloudinary (commented):
Hi @gabrielcld2, this looks good, but I was wondering if we could also add if statements before using `sync_types` in the `class-sync.php` method (especially `get_signature`). I'd like to ensure there are no warrnings/errors in the future with that property.

gabrielcld2:
Hi @aatanasov-cloudinary - Sure, better be safe just in case! I've added some extra checks for all cases where `sync_types` is empty https://github.com/cloudinary/cloudinary_wordpress/pull/1085/commits/69b3bf407514338e656fa26e95856fcf9561619f 

This should avoid any issues in the future if this property is somehow set to null elsewhere. Let me know your thoughts!

aatanasov-cloudinary (approved):
Looks good!

PR #1088, opened 2025-08-29
Avoid SVG conversion for pulled SVG from Cloudinary if SVG support is enabled

gabrielcld2:
When we pull an SVG from Cloudinary, if SVG support is enabled we should expect it to be disabled as an SVG file. 

However, it is currently displaying as a PNG.

## Approach

- If the setting is supported, disable SVG conversion using the `cloudinary_convert_media_types` filter

## QA notes

-  Go to Cloudinary Image Settings page
   - Make sure **SVG Support** toggle is enabled
- Create a new post/page
  - Add an image block
    - Import an SVG from Cloudinary (if none are present in your account, import one using Cloudinary DAM)
  - Make sure the SVG is displayed as an SVG on the editor
  - Publish and view the post 
    - Check that it displays as an SVG on the frontend too and that it has expected SVG transformations (e.g. `f_svg,q_auto/fl_sanitize`)
- Go back to Cloudinary Image Settings page
   - Toggle OFF the **SVG Support** setting
- Follow the same previous step to add a new post with an uploaded SVG
  - It should be converted to a PNG

### DIFF
--- a/php/class-svg.php
+++ b/php/class-svg.php
@@ -218,6 +218,21 @@ class SVG extends Delivery_Feature {
 	}
 
 	/**
+	 * Disable SVG conversion.
+	 *
+	 * @param array $base_types The base types for conversion.
+	 *
+	 * @return array
+	 */
+	public function disable_svg_conversion( $base_types ) {
+		if ( $this->is_active() && ! empty( $base_types['svg'] ) ) {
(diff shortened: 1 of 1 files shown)


aatanasov-cloudinary (approved):
@gabriel-detassigny, looks good!

PR #1130, opened 2026-01-27
Add support for background images within Elementor

gabrielcld2:
This ensures that background images generated within Elementor's CSS files are handled accordingly.

## Approach

- Hook into the Elementor parse_css hook. 
  - Check for any background image elementor keys
  - Generate the corresponding cloudinary URL for the media
  - Inject the right CSS rule to override the default one within Elementor
- Flush the elementor cache whenever a cache flush occurs on Cloudinary
  - This ensures any setting change is directly applied for Elementor background images

## QA notes

- Activate the Elementor plugin
  - Create a page with Elementor
    - Within the elementor page, add an element like a heading
    - Edit the heading, go to _Advanced_ and open the **background** section
    - Apply a background image (and if you want, you can also apply a hover background image, or even some viewport specific ones like tablet or mobile background images)
  - Publish and view the page
  - Inspect the element and check its CSS contains the cloudinary URL for the background image
  - Go to Cloudinary -> Image Settings
    - Make some changes for instance the image format
    - Save the changes
    - Reload the elementor page and make sure the changes applied to the background image
- Deactivate the Elementor plugin
  - Make sure the Cloudinary plugin still works, and doesn't expect Elementor to be enabled

### DIFF
--- a/php/class-plugin.php
+++ b/php/class-plugin.php
@@ -14,6 +14,7 @@ use Cloudinary\Component\Setup;
 use Cloudinary\Delivery\Lazy_Load;
 use Cloudinary\Delivery\Responsive_Breakpoints;
 use Cloudinary\Assets as CLD_Assets;
+use Cloudinary\Integrations\Elementor;
 use Cloudinary\Integrations\WPML;
 use Cloudinary\Media\Gallery;
 use Cloudinary\Sync\Storage;
@@ -31,7 +32,7 @@ final class Plugin {
 	 *
 	 * @since   0.1
 	 *
-	 * @var     Admin|CLD_Assets|Connect|Dashboard|Deactivation|Delivery|Extensions|Gallery|Lazy_Load|Media|Meta_Box|Relate|Report|Responsive_Breakpoints|REST_API|State|Storage|SVG|Sync|URL[]|WPML|null
+	 * @var     Admin|CLD_Assets|Connect|Dashboard|Deactivation|Delivery|Extensions|Gallery|Lazy_Load|Media|Meta_Box|Relate|Report|Responsive_Breakpoints|REST_API|State|Storage|SVG|Sync|URL[]|WPML|Elementor|null
 	 */
 	public $components;
 	/**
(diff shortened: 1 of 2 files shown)


aatanasov-cloudinary (commented):
Hi @gabriel-detassigny, I tested the PR on my side by adding a background image to a container element, and the local URL is used. The heading element works.

**Back-end**

Image

**Front-end**

Image

I confirm the `replace_background_images_in_css` hook was triggered on save (added a log to test it).

gabrielcld2:
Thanks @aatanasov-cloudinary ! It turns out, the container elements used a slight different syntax for its keys (heading and such use `_background_image` while the container types use `background_image`). 

Furthermore, the container applies an extra suffix internally which I couldn't retrieve directly via Elementor's internals, so I've applied it directly in that case. This fixes the issue now.

I've also added some extra defensive code to check all Elementor methods interactions, as we can't be certain these methods will still exist in future versions. 
Another option would be to use
```php
try {
 // internal method calls.
}
catch ( \Throwable $e ) {
}
```
however `Throwable` isn't available in PHP 5.6 so perhaps this would be a good future improvement once we drop PHP 5.6 support.

Anyway, have a look and let me know your thoughts!

aatanasov-cloudinary:
@gabriel-detassigny, thanks, when adding a background image to a container, I can see we have the following in the CSS:

Screenshot 2026-01-30 at 14 17 46

Can you please check if we can optimize this as well (both selectors use background images delivered by Cloudinary)? 

Also, it will be helpful to look into the background overlay case. It uses `::before`:

**Front-end**

Screenshot 2026-01-30 at 14 18 28

**CMS**

Screenshot 2026-01-30 at 14 20 20

gabrielcld2:
Thanks @aatanasov-cloudinary 
For extra context: the rule you found with the URL still set as the local one was actually the remaining default Elementor one. When we add our own rules, they either:
- replace the existing one if they exactly match the same selector
- get added as a new CSS rule

So I've made changes to try to ensure we replace the existing rules (including the overlay which I've added support for). It's not an easy job though as Elementor checks for a lot of pseudo-elements, child elements, etc. And we can't guarantee this won't change in a future version.

Still, it seems to work well right now, so have a look and let me know!

aatanasov-cloudinary (approved):
@gabrielcld2, thanks! The background images and overlays are delivered from Cloudinary when testing containers.

PR #1132, opened 2026-01-28
Optimize the lazy-loading DPR numbers

aatanasov-cloudinary:
### Description

In some cases, DPR might look like those numbers:
 - 1.6666666666666667
 - 1.1041666269302368 

Which generates a lot of asset versions. 

### Approach

We round the number, and the step is 0.5, and the closest one will be used. Examples:
 - 1.6666666666666667 => 1.5
 - 1.1041666269302368 => 1 
 - 1.766766 => 2.0
 - 1.976 => 2.0
 - 1.33333213 => 1.5

## QA notes

- Enable the Lazy Load feature.
- Select the `Auto 2x` option from the `DPR Settings`.
- Go to a page with images delivered by Cloudinary.
- Zoom in/out on your browser (refresh the page before each test).
- Ensure the DPR in the URL is rounded (e.g 1.0, 1.5, 2.0).


src/js/inline-loader.js
```diff
@@ -151,11 +151,20 @@ const CloudinaryLoader = {
...
+		}
+
 		if (
 			'max' !== maxDensity &&
 			'auto' !== deviceDensity
 		) {
 			maxDensity = parseFloat( maxDensity );
+
+			// Round maxDensity to nearest 0.5 to maintain consistency
+			maxDensity = Math.round( maxDensity * 2 ) / 2;
```
gabrielcld2: **suggestion (non-blocking):** I wonder if we could introduce a simple utility function to do that rounding. I know it's one line right, but if we ever change the rounding logic, it would be nice to have to do it in one place. 

Not a blocker though as the current approach works fine 🙂

gabrielcld2 (approved):

src/js/inline-loader.js
aatanasov-cloudinary: Thanks @gabrielcld2 , prepared a utility method :)

PR #1133, opened 2026-01-28
added default 21

atardadi:
## Approach

- Max images recommends a value between three and 40, but the default value that I got was 10. This is not intuitive. It needs to be the average of the recommended value. 

## QA notes
- Make sure that the default value is actually pre-populated when connecting to the plugin.


gabrielcld2:
Thank you for your contribution, however we prefer to keep the default value at 10 in this specific instance. This seems a more reasonable amount of breakpoints by default.

PR #1160, opened 2026-03-30
Fix CSS & JS deprecation warnings following node upgrade

gabrielcld2:
Due to the recent node upgrade, several deprecation warnings were being shown.
These are now addressed here.

## Approach

- Replace `render` with `createRoot`
- Migrate SCSS `@import` to either `@use` or `@forward`
- Replace SCSS deprecated color functions

Note: there are two warnings left due to file sizes. These might have to remain for now:
1. The cloudinary CSS file was already quite large. Splitting it into smaller files might be a bit more complicated.
2. `syntax-highlight.js` is large due to the CodeMirror dependency. That file is thankfully only enqueued on the admin page requiring it only, so perhaps we can leave with this for now. 

## QA notes

- Run `npm run build` to make sure assets are still building and not showing warnings (except for the two file sizes mentioned above)
- Do some regression testing on Cloudinary settings pages, the editor, and the frontend.


PatelUtkarsh (approved):
LGTM! 

Build is clean, all Sass warnings are gone. Tested locally and everything looks good.

PR #1172, opened 2026-05-12
Release automation

gabrielcld2:
## Deployment process

This is what the production release process will look like from now on:

### 1. Prepare master branch

This is pretty much as before:

- prepare all changes in a PR against `master`
- this should also include the following:
  - updated version in both `.version` and `package.json`
  - updated `readme.txt` changelog
- merge the master PR

### 2 Create and publish GH release

This is also similar to what was done before:

- Go to the repository on GitHub → Releases → Draft a new release
  - Create a new tag (e.g. 3.3.4) targeting `master`
  - Set the release title and notes (copy and format the changelog from the `readme.txt`) 
  - Publish as a standard release
   - Note: pre-release can be used to do a dry-run instead which will trigger the workflow but won't push to wp.org's SVN repository.

### 3. The GH workflow

Publishing the release will automatically trigger `.github/workflows/deploy-to-wp-org.yml`, which:

1. Runs the full CI suite to ensure the build isn't broken
2. Installs npm and Composer dependencies
3. Builds JS/CSS assets (`npm run build`)
4. Copies all plugin files into `build/` and update version references (via `grunt prepare`)
5. Deploys to WP.org SVN via the 10up action — commits to `trunk/`, creates `tags/<version>/`, and updates `assets/`
6. Attaches a .zip of the plugin to the GitHub release

## QA notes

- We can try triggering a dry-run by creating a pre-release in GH.


.github/workflows/deploy-to-wp-org.yml
```diff
@@ -0,0 +1,107 @@
...
+        env:
+          BUILD_DIR: 'build'
+          SLUG: 'cloudinary-image-management-and-manipulation-in-the-cloud-cdn'
+          # Use secrets to authenticate with WP.org.
+          SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
+          SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
+
+      # After the deployment, we also want to create a zip and upload it to the release on GitHub.
+      - name: Upload release asset
+        uses: actions/upload-release-asset@v1
```
PatelUtkarsh: **💡 thought:** This action is archived and deprecated by GitHub since 2021.

Should we migrate it to more active ones like: 


```
- name: Upload release asset
  uses: softprops/action-gh-release@v2
  with:
    files: ${{ steps.deploy.outputs.zip-path }}
```

.github/workflows/deploy-to-wp-org.yml
```diff
@@ -0,0 +1,107 @@
...
+      - name: Cache Composer packages
+        uses: actions/cache@v4
+        with:
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: ${{ runner.os }}-php-${{ hashFiles( 'composer.lock' ) }}
+          restore-keys: ${{ runner.os }}-php-
+
+      # Install dependencies.
+      - name: Install NPM dependencies
+        run: npm install
```
PatelUtkarsh: **💬 suggestion:** In CI we use `npm ci ` we can keep it same here.

```suggestion
        run: npm ci
```

.github/workflows/deploy-to-wp-org.yml
```diff
@@ -0,0 +1,107 @@
...
+
+      - name: Install Composer dependencies
+        run: composer install --no-dev
+
+      # Build.
+      - name: Build
+        run: npm run build
```
PatelUtkarsh, line 76: **💬 suggestion:** The release pipeline trusts that the human releaser bumped .version, package.json, and readme.txt to match the tag they're publishing. There's nothing here to fail the build if git tag == "3.3.4" but .version == "3.3.3". Given the PR explicitly notes the human has to remember to update both files, a 5-line sanity check is cheap insurance:
```
- name: Verify version matches tag
  run: |
    TAG="${GITHUB_REF_NAME#v}"
    FILE_VERSION=$(cat .version | tr -d '[:space:]')
    if [ "$TAG" != "$FILE_VERSION" ]; then
      echo "::error::Tag $TAG does not match .version $FILE_VERSION"
      exit 1
    fi
```

.github/workflows/ci.yml
```diff
@@ -2,6 +2,7 @@ name: CI
 
 on:
   push:
+  workflow_call:
   pull_request:
```
PatelUtkarsh, line 7: **💬 suggestion:** With `push:` and `pull_request:` and `workflow_call:` here every PR push fires CI twice (once for the push event on the branch, once for the PR). This isn't introduced by this PR but the `workflow_call` addition is a good moment to audit. Suggestion (out of scope, but worth a follow-up): scope `push:` to `branches: [develop, master]`.

.github/workflows/deploy-to-wp-org.yml
```diff
@@ -0,0 +1,107 @@
...
+  lint_and_test:
+    uses: ./.github/workflows/ci.yml
+
+  deploy_to_wp_repository:
+    needs: lint_and_test
+    name: Deploy to WP.org
+    runs-on: ubuntu-22.04
+    permissions:
+      contents: write
+      packages: read
```
PatelUtkarsh: **💬 suggestion:** Is this required? if not we can clean it up.

PatelUtkarsh (approved):
Left some suggestions but overall looks good!

.github/workflows/deploy-to-wp-org.yml
gabrielcld2: @PatelUtkarsh It seems that "softprops/action-gh-release" serves a different purpose from my understanding though, no? 🤔 It seems to be aimed at creating github releases, while here we're trying to upload assets to the manually created release https://github.com/softprops/action-gh-release

.github/workflows/deploy-to-wp-org.yml
gabrielcld2: done 👍 I've also changed to use ubuntu-latest so that it's consistent with the CI workflow [de05f63](https://github.com/cloudinary/cloudinary_wordpress/pull/1172/commits/de05f6395c74d6efee8605cdf9c86459f0a12c52)

.github/workflows/ci.yml
gabrielcld2: Done [132618d](https://github.com/cloudinary/cloudinary_wordpress/pull/1172/commits/132618d5910d4f6e61ae50705159fdabe0eb4eff)

.github/workflows/deploy-to-wp-org.yml
gabrielcld2: Good idea 👍  [e5456e6](https://github.com/cloudinary/cloudinary_wordpress/pull/1172/commits/e5456e61041cc2040c091f192ef760b4fcadd9e2) Done here

.github/workflows/deploy-to-wp-org.yml
gabrielcld2: Yes, that makes sense! done [cf14f66](https://github.com/cloudinary/cloudinary_wordpress/pull/1172/commits/cf14f66e008f4ec1e2de25d99f890ca7b400672b)

.github/workflows/deploy-to-wp-org.yml
utkarshcloudinary: @gabrielcld2 Good question, but `softprops/action-gh-release` actually handles this case too 👍

When the workflow is triggered by a `release` event, the action **does not create a new release** - it detects the existing release for the current tag and updates it in place, uploading anything passed via `files:` as release assets. Creating releases is just one of its modes; attaching assets to an existing release is the most common usage and exactly what the (archived) `actions/upload-release-asset` did.

From their README ([Uploading release assets](https://github.com/softprops/action-gh-release#%EF%B8%8F-uploading-release-assets)):

> "If a tag already has a GitHub release, the existing release will be updated with the release assets."

The migration would be a drop-in replacement:

```yaml
# After the deployment, upload the zip to the existing GitHub release.
- name: Upload release asset
  uses: softprops/action-gh-release@v2
  with:
    files: ${{ steps.deploy.outputs.zip-path }}
```

No `upload_url`, `asset_name`, or `asset_content_type` plumbing needed — it infers all of that. The `GITHUB_TOKEN` env is also unnecessary (it uses the default token), though the job may need `permissions: contents: write` depending on the repo's default workflow permissions.

Two things to be aware of:
1. The asset will be named after the actual zip filename rather than `${{ github.event.repository.name }}.zip` if keeping the old name matters, add a rename step before upload.
2. If the repo's default `GITHUB_TOKEN` permissions are read-only, the job needs an explicit `permissions: contents: write`.

Since `actions/upload-release-asset` has been archived since 2021 and won't receive security or Node-runtime updates, I'd still suggest making the switch in this PR while we're already touching this workflow.

.github/workflows/deploy-to-wp-org.yml
```diff
@@ -0,0 +1,114 @@
...
+  # the "published" type will trigger. Since we want a workflow to run when stable and pre-releases
+  # publish, we subscribe to "published" instead of "released" and "prereleased".
+  #
+  # See: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release
+  release:
+    types: [ published ]
+
+jobs:
+  lint_and_test:
+    uses: ./.github/workflows/ci.yml
```
utkarshcloudinary, line 30: With workflow_call, secrets are not passed automatically; the caller must add secrets: inherit (or `ci.yml` must declare the secret under `on.workflow_call.secrets`). As written, every release-triggered run will execute E2E with an empty `CLOUDINARY_E2E_URL`, which will silently test the wrong target or fail; blocking the deploy since `deploy_to_wp_repository` has `needs: lint_and_test`.

```suggestion
    uses: ./.github/workflows/ci.yml
    secrets: inherit
```

utkarshcloudinary (approved):

.github/workflows/deploy-to-wp-org.yml
gabrielcld2: Good catch! done here [0682ff8](https://github.com/cloudinary/cloudinary_wordpress/pull/1172/commits/0682ff8d78ddc430eb4c71a08c9d4fcac5ab1251)

.github/workflows/deploy-to-wp-org.yml
gabrielcld2: 👍 Got it. that makes sense. implemented here [6684b8d](https://github.com/cloudinary/cloudinary_wordpress/pull/1172/commits/6684b8d9a5e47a030a94b98a5ca370309ef01140)

utkarshcloudinary (approved):
Looks good!

PR #1174, opened 2026-05-18
Add collapsible behavior to term edit page

gabrielcld2:
## Approach

  - Refactored `taxonomies.js` to support multiple collapsible instances per page using `querySelectorAll`, resolving content targets via a `data-collapsible-target` attribute and `getElementById`. Visibility is now toggled via
  the `hidden` attribute (replacing `style.display`) to correctly handle both `<div>` and `<tbody>` elements. Arrow icon state uses `classList.toggle` with a force argument.
  - Added `aria-expanded`, `aria-controls`, and `aria-label` to all collapsible toggle buttons for screen reader accessibility.
  - Extracted the card box-model styles from `.cloudinary-collapsible` into a `--card` BEM modifier in `_settings.scss`, keeping layout-specific styles separate from the shared toggle/content rules. Extended the `__toggle`
  heading selector to cover both `h2` and `h3`.
  - Updated `taxonomy-transformation-fields.php` to use the `--card` modifier, wire up `data-collapsible-target` / `id` on the toggle and content elements, and replace `style="display:none"` with the `hidden` attribute.
  - Added collapsible behavior to `taxonomy-term-transformation-fields.php` using paired `<tbody>` elements per transformation context — a *toggle row* containing the heading and button, and a *content* `<tbody>` (initially
  `hidden`) wrapping the existing field rows — preserving valid table structure throughout.
- Added custom CSS class `cloudinary-term-transformations` to the `tbody` elements added by Cloudinary so that it can be targeted for custom styling if needed.

## QA notes

- Go to the taxonomy term edit form
  - Confirm the custom cloudinary fields are now shown behind a dropdown
  - You should be able to toggle the dropdown to see its content
- Confirm the the taxonomy term creation form still shows the dropdown as before


src/js/components/taxonomies.js
```diff
@@ -1,29 +1,36 @@
-const toggler = document.querySelector( '.cloudinary-collapsible__toggle' );
+const togglers = document.querySelectorAll( '.cloudinary-collapsible__toggle' );
```
PatelUtkarsh, line 1: **💡 thought:** The listener is on the wrapper `.cloudinary-collapsible__toggle` (a `<div>`). The `<button>` inside fires the click which bubbles up — that works, but:
- The heading `<h3>/<h2>` is also inside the wrapper and becomes clickable without any keyboard equivalent or role / visible affordance.
- The previous CSS provided cursor: pointer on &__toggle to telegraph clickability; this PR removes that line, so the wrapper is now silently clickable. Either restore cursor: pointer on the wrapper, or scope the listener to the button only.

PatelUtkarsh (approved):
Looks good! Left 1 suggestion.

src/js/components/taxonomies.js
gabrielcld2: thanks @PatelUtkarsh ! I've restored the `cursor: pointer` on the wrapper now [b56ae6a](https://github.com/cloudinary/cloudinary_wordpress/pull/1174/commits/b56ae6a52add41b3a6c72819aa34f4bb369b7d11) Please have another look!

PatelUtkarsh (approved):
Looks good!

PR #1177, opened 2026-06-03
test(e2e): verify Cloudinary image delivery on front-end

utkarshcloudinary:
## Approach

Adds a Playwright e2e test that verifies images on the WordPress front-end are served from Cloudinary once the plugin is configured with valid credentials.

**New files**
- `tests/e2e/cloudinary-image-delivery.spec.js` — the spec.
- `tests/e2e/utils/connection.js` — helper that connects the plugin via WP-CLI (`cloudinary_connect` option) and parses the cloud name out of `CLOUDINARY_E2E_URL`.
- `tests/e2e/fixtures/test-image.jpg` — small (~2KB) JPEG fixture.

**What the test does**
1. `beforeAll`: connect the plugin via WP-CLI using `CLOUDINARY_E2E_URL` (no UI walk; that path is covered by the wizard spec).
2. `beforeEach`: upload the fixture via the WP REST media endpoint, create a published post that uses the attachment as both the featured image and an inline image block, then run `wp cloudinary sync` so the asset is pushed to Cloudinary deterministically.
3. Visit the post on the front-end, scroll each image into view (the wizard enables lazy-load, which means `src` starts as an SVG placeholder until JS replaces it), then assert that the resolved `src` (and the first `srcset` candidate when present) for both the featured image and the inline image have host `res.cloudinary.com` and a pathname starting with `/<cloud_name>/`.
4. `afterEach`: delete the post and attachment via WP-CLI.

**Decisions worth flagging**
- The inline-image selector is scoped to `.wp-block-image` because Twenty Twenty-Five tags the featured image with `wp-image-<ID>` too.
- Cleanup uses WP-CLI rather than the REST API because the wp-env test site does not use pretty permalinks, which makes appending `force=true` to the `?rest_route=`-style REST URL fragile.
- No assertions on specific transformation strings (`f_auto`, `q_auto`, etc.). The intent is to catch a regression where URLs stop being rewritten at all; transformation specifics are an implementation detail of the plugin and likely to evolve.

## QA notes

Requires `CLOUDINARY_E2E_URL` to be set (already wired via `.env` and the e2e workflow secret).

```
npm run test:e2e -- tests/e2e/cloudinary-image-delivery.spec.js
```

Expected: 1 passed in ~13s. Full suite (`npm run test:e2e`) should report 8 passed.

To inspect the run:
```
npm run test:e2e:debug -- tests/e2e/cloudinary-image-delivery.spec.js
```


utkarshcloudinary:
@gabrielcld2 This is ready for review.

gabrielcld2 (approved):

PR #1180, opened 2026-06-18
Add PHPStan static analysis at level 0

utkarshcloudinary:
Adds PHPStan static analysis to the project.

## Approach

- Set up **PHPStan 2.x** with `szepeviktor/phpstan-wordpress` at **level 0**, run via `composer phpstan`.
- Added stub packages (`php-stubs/wp-cli-stubs`, `php-stubs/woocommerce-stubs`) and minimal hand-written stubs under `tests/phpstan/stubs/` for the optional WPML and WordPress VIP symbols the plugin references (no official stub packages exist for those).
- **Fixed all level-0 findings in code rather than baselining them.** Notable real bugs surfaced:
  - Filters mistakenly registered with `add_action`, so their return value was silently discarded: `wp_resource_hints`, `http_request_args`, `template_include` (broke the debug-view template swap), `cloudinary_build_queue_query`, `cloudinary_thread_queue_details_query`.
  - Incorrect `$accepted_args` counts on several hook registrations.
  - Class-case bugs: `SYNC` → `Sync`, `media_status` → `Media_Status`.
  - Missing return paths in `File_System::get_file_src_root()` and `Component::render()`.
  - `$old_meta` could be `false` before `unset`; removed a redundant dead-code `unset` of `srcset`.
  - Replaced `defined( 'DOING_AJAX' ) && DOING_AJAX` with `wp_doing_ajax()`.
  - Normalized non-standard hook docblocks (`@param $name {type}`) to valid PHPDoc and added `@method` annotations for the `Component` magic methods.
- The only suppressed check is `HookCallbackRule`'s "action callback should not return anything", which is too strict for dual-purpose methods that are both called directly and registered on a real action; its argument-count checks remain active.
- **CI**: added a dedicated `PHPStan` job (PHP 8.3, runs once) and cached PHPStan's result cache for fast subsequent runs (~17s cold → <1s warm).
- `phpcs` excludes `tests/phpstan/stubs/` since stub files intentionally mirror third-party API shapes.

## QA notes

1. Run `composer install` then `composer phpstan` — it should report **`[OK] No errors`**.
2. Confirm the new **PHPStan** check runs and passes in CI on this PR.
3. Verify the result cache speeds up a second local run (`composer phpstan` again should complete in well under a second).
4. Smoke-test the behavioral fixes:
   - DNS prefetch hints (`wp_resource_hints`) still output on the front end.
   - With `CLD_DEBUG` enabled, the debug view template replacement still works (`template_include`).
   - Sync queue building and thread-queue queries still include the plugin's post type.

### DIFF
--- a/php/cache/class-cache-point.php
+++ b/php/cache/class-cache-point.php
@@ -107,9 +107,9 @@ class Cache_Point {
 		 * @hook    cloudinary_on_demand_sync_limit
 		 * @default 100
 		 *
-		 * @param $value {int} The default number of static assets.
+		 * @param int $value The default number of static assets.
 		 *
-		 * @return {int}
+		 * @return int
 		 */
 		$this->sync_limit = apply_filters( 'cloudinary_on_demand_sync_limit', 100 );
 		$this->register_post_type();
@@ -118,7 +118,7 @@ class Cache_Point {
 		add_filter( 'get_post_metadata', array( $this, 'get_meta' ), 10, 3 );
 		add_filter( 'delete_post_metadata', array( $this, 'delete_meta' ), 10, 4 );
 		add_action( 'shutdown', array( $this, 'meta_updates' ) );
-		add_action( 'wp_resource_hints', array( $this, 'dns_prefetch' ), 10, 2 );
+		add_filter( 'wp_resource_hints', array( $this, 'dns_prefetch' ), 10, 2 );
 	}
 
 	/**
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,6 +66,15 @@ jobs:
       - name: Install Composer dependencies
         run: composer install --no-interaction --no-progress
 
+      - name: Cache PHPStan result cache
+        uses: actions/cache@v4
+        with:
+          path: .phpstan-cache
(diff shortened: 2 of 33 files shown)


php/cache/class-cache-point.php
```diff
@@ -107,9 +107,9 @@ public function __construct( Cache $cache ) {
 		 * @hook    cloudinary_on_demand_sync_limit
 		 * @default 100
 		 *
-		 * @param $value {int} The default number of static assets.
+		 * @param int $value The default number of static assets.
 		 *
-		 * @return {int}
+		 * @return int
```
gabrielcld2, line 112: **issue:** Note this has the potential to break the auto-generated docs https://cloudinary.github.io/cloudinary_wordpress/ Perhaps something to double-check! The auto-generated tool might still work with this new syntax. Else, we can maybe see if this can be ignored in PHPStan config?

gabrielcld2 (approved):

php/cache/class-cache-point.php
gabrielcld2: Note: this was fixed in the dedicated documentation branch https://github.com/cloudinary/cloudinary_wordpress/pull/1191

PR #1181, opened 2026-06-22
Add filter to disable Elementor background images integration

gabrielcld2:
## Approach

- Some specific features in Elementor Pro have issues with our attempt at replacing background images URLs
- As a starting point, we'll give the option to disable that specific integration conditionally

## QA notes

- Install the **Elementor** plugin
- Create a page with **Elementor**
  - Within the elementor page, add an element like a heading
  - Edit the heading, go to _Advanced_ and open the background section
  - Apply a _background image_
  - Save the page
- Publish and view the page
  - Inspect the element and check its CSS uses the **cloudinary CDN URL** for the background image
- Implement a filter to disable this within your theme or a plugin:
  - `add_filter( 'cloudinary_elementor_replace_background_images_in_css', fn() => false );`
- Force both Cloudinary and Elementor cache to be flushed by going to to Cloudinary image settings page and saving changes
- Re-inspect the element on the page and confirm it's now being served using the **WordPress URL**


utkarshcloudinary (approved):
Looks good!

PR #1182, opened 2026-06-22
Fix duplicate assets on DB write fails scenarios

gabrielcld2:
When Cloudinary returns `existing: true` (overwrite=false, asset already exists), the plugin retried with a unique suffix, creating a new Cloudinary asset. If the server killed the PHP process after the upload but before saving `_public_id`, WordPress never recorded the asset. Each subsequent autosync cycle repeated this, producing one new Cloudinary duplicate per cycle (102 in the reported case).

When WordPress has no `_public_id` for an attachment, the conflicting Cloudinary asset is from a previously failed upload attempt. Overwriting it is safe and avoids creating another suffixed duplicate.

## Approach

- In the specific scenario detailed above, we tell Cloudinary to overwrite the assets on upload instead of uploading a new assets

### Steps to reproduce

- Upload a new image in the media library
  - wait a few seconds until it's synced to Cloudinary
  - Make note of the attachment ID
- In your terminal run the following commands (replace the **123** IDs with your attachement ID):
  - Replicate the stuck step: `npx wp-env run cli wp eval '$m = get_post_meta(123, "_cloudinary", true); $m["_public_id"] = ""; $m["_sync_signature"] = []; update_post_meta(123, "_cloudinary", $m); echo "Done\n";'`
  - Trigger the upload: `npx wp-env run cli wp eval 'Cloudinary\get_plugin_instance()->get_component("sync")->managers["push"]->process_assets(123);'`
  - Check the saved public_id: `npx wp-env run cli wp eval 'echo get_post_meta(123, "_cloudinary", true)["_public_id"] . "\n";'`

**On `master` branch**: the public_id will have a suffix like _123_xxxxx, meaning the asset is duplicated within Cloudinary's backend. 

**On this branch**: the original public_id won't have a suffix, indicating that it was overwritten instead of duplicated.

### DIFF
--- a/php/sync/class-upload-sync.php
+++ b/php/sync/class-upload-sync.php
@@ -298,10 +298,11 @@ class Upload_Sync {
 	 * @param int         $attachment_id The attachment ID.
 	 * @param string|null $type          Optional Sync type.
 	 * @param string|null $suffix        An optional suffix.
+	 * @param bool        $overwrite     Whether to overwrite an existing Cloudinary asset.
 	 *
 	 * @return array|\WP_Error
 	 */
-	public function upload_asset( $attachment_id, $type = null, $suffix = null ) {
+	public function upload_asset( $attachment_id, $type = null, $suffix = null, $overwrite = false ) {
 
 		add_filter( 'cloudinary_doing_upload', '__return_true' );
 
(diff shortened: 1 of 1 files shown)


utkarshcloudinary (approved):
Looks good!

PR #1191, opened 2026-06-30
build(docs): teach wp-hookdoc to read standard PHPDoc

utkarshcloudinary:
Relates to the PHPStan work in #1180.

When #1180 normalizes the `@hook` docblocks to standard, PHPStan-clean PHPDoc (`@param Type $name` / `@return Type`), the documentation generator breaks. The docs are built with `wp-hookdoc`, a JSDoc plugin that expects the type in `{braces}` (`@param {Type} $name`). With valid PHPDoc the hook pages render the type where the variable name should go, drop the type metadata, and leak the variable name into the description. This PR fixes the docs pipeline so the PHP source can keep valid PHPDoc.

## Approach

- Patch `wp-hookdoc@0.2.0` (via `patch-package`) so its `beforeParse` hook normalizes standard PHPDoc tag syntax to the JSDoc syntax the parser expects, **only inside `@hook` blocks**:
  - `@param Type $name Desc` (standard PHPDoc) → `@param {Type} $name Desc`
  - `@param $name {Type} Desc` (legacy hookdoc) → `@param {Type} $name Desc`
  - `@return Type Desc` → `@return {Type} Desc`
  - Tags already using the `{Type}` form are left untouched.
- Handle PHPDoc fully-qualified class names that JSDoc's type parser rejects: leading-backslash FQCNs like `\WP_Post` / `\Cloudinary\Sync` are normalized to `WP_Post` / `Cloudinary.Sync` so they no longer error during the build.
- Wire `patch-package` into `postinstall` (`patch-package && composer install`) and add it as a devDependency; the committed patch lives in `patches/wp-hookdoc+0.2.0.patch`.
- Regenerate `package-lock.json` so `npm ci` is in sync (reconciles pre-existing transitive drift, e.g. `date-fns` 4.1.0 → 4.4.0, and adds `patch-package`).

This keeps the PHP source as valid PHPDoc (PHPStan stays green) while the generated hook documentation continues to render correctly.

## QA notes

1. `npm install` — confirm `patch-package` applies the patch on postinstall (look for `wp-hookdoc@0.2.0 ✔`).
2. `npm ci` — completes without "package.json and package-lock.json are in sync" errors.
3. `npm run build:docs` — completes with **no** `Unable to parse a tag's type expression` / `invalid name or invalid syntax` errors.
4. Spot-check a few generated hook pages (e.g. `docs/cloudinary_dns_prefetch_types.html`, `docs/cloudinary_can_sync_asset.html`, `docs/cloudinary_context_options.html`): the filter signature shows the variable name and `→ {Type}` return, the Parameters table lists `$name`, its type, and description correctly, including union (`bool|WP_Error`) and FQCN (`WP_Post`, `Cloudinary.Sync`) types.
5. Verified end-to-end against the level-1 PHPStan branch (a superset of #1180): docs build cleanly from the converted, valid-PHPDoc source.


package-lock.json
PatelUtkarsh, line 1: Needed to re-generate the lock file.

gabrielcld2 (approved):
Looks good 👍 tested locally

PR #1226, opened 2026-08-03
Instrument connection, sync, settings, media, cache, features, and deactivation analytics events

gabrielcld2:
## Approach

Wires the remaining 7 event categories from the Analytics Event Tracking Spec POC's custom-events framework:
**connection management, asset sync, settings & navigation, media & asset actions, non-media cache, extensions & gallery, and deactivation** (30 events total).

No changes to the transport layer (`Analytics::track()` / `Analytics.track()`) — this is call-site wiring only.

## Design decisions worth a second look

- **`bulk_sync_started`** only covers the manual REST trigger (`Push_Sync::rest_start_sync()`). The `auto_restart` case inside `Sync_Queue::stop_maybe()` is treated as internal plumbing and isn't instrumented.
- **`asset_sync_failed`** hooks the broad, already-generic `Sync::log_sync_result()` rather than the narrower`Upload_Sync::upload_asset()`, trading a small `asset_type` lookup cost for covering every sync type in one place.
- **`sync_completed`** needed new state (`Sync_Queue::mark_run_started()` / `tally_run_result()` / `track_run_completed()`) since no
  existing signal survives the queue's internal restart cycles.
- **`account_switched`** is new comparison logic in `Connect::verify_connection()` — `Connect::switch_account()` turned out to be dead code (no callers since 2021), not something to hook into.
- **Cache-category events target `php/assets/class-rest-assets.php`, not `php/class-cache.php`** — the latter is never instantiated (dead code), confirmed via exploration before wiring anything.
- **`gallery_configured`**'s `layout`/`media_count` are parsed out of the `gallery_config` field's serialized JSON blob rather than being real submitted fields.

## Unrelated fix included: duplicate page-view bug

While QA'ing `settings_page_viewed`, we found it firing twice per single page load on Image Settings / Video Settings. Root cause: `Image_Preview::preview()` set the sample ``'s `src="#"`, which resolves to the *current page's own URL* — so the browser fired a second real request back at the same admin page (as an image fetch) before `global-transformations.js` replaced it with a real preview URL. `Video_Preview` and `Breakpoints_Preview` both extend `Image_Preview` and inherited the same bug.

This is a genuine, pre-existing perf bug unrelated to the analytics work, but it directly corrupted the accuracy of the new `settings_page_viewed` event on exactly the pages it affects, so it's fixed here rather than filed separately: `src="#"` → an inert 1x1 data-URI placeholder. Confirmed via raw Apache access logs (down to one request) and the capture log (down to one event) that the duplicate is gone; full Playwright suite still green afterward.

## Unrelated fix included: `attributes:class` never applied on the `link` UI component

`special_offer_clicked` needed the special-offer `<a>` to render with a stable class, set via `attributes:class` on the `link` component's config. It never actually applied: `Params_Trait::get_param()` splits its path on `.`, so a colon-delimited key never matched the nested `attributes` array. Fixed in `php/ui/component/class-link.php` to read `attributes` directly. Confirmed no other caller relied on the broken fallback.

### Automation

**28 of the 30 events in this PR are covered by automated e2e specs** — run `npm run test:e2e` (one spec file per category: `tests/e2e/{connection,sync,settings,media,cache,features,deactivation}-analytics.spec.js`) and confirm it's green. That's the QA gate for everything except the two events below, which have no automatable trigger:

- **`connectivity_check_failed`** — cron-only (daily ping / REST-API self-check), no UI path exists. To check manually: connect with a valid Cloudinary URL, then break it (`wp option update cloudinary_connect '{"cloudinary_url":"cloudinary://invalid"}' --format=json`), run `wp cron event run cloudinary_status --due-now`, and confirm the event in the capture log (`.wp-env/mu-plugins/analytics-capture.php`, `wp cloudinary analytics-events`).
- **`account_switched`** — needs two real Cloudinary cloud names. Connect with account A's credentials via the wizard or Connect settings page, then submit account B's credentials on the same page; confirm the event carries `previous_cloud_name`/`new_cloud_name`.

**Regression check:** confirm no PHP notices/fatals in `wp-content/debug.log` and no new JS console errors while the suite runs — none of this should be visible to a site admin who isn't looking for it.

#### Setup (one-time per session)

1. Start the environment: `npm run env:start`
Note: The `.wp-env/mu-plugins/analytics-capture.php` mu-plugin auto-loads and intercepts every outgoing call to `analytics-api.cloudinary.com`, logging the payload instead of sending it (and blocking the real send).
2. Make sure the site is "connected" to Cloudinary via real credentials

#### Per-event loop

3. Clear the capture log right before the action:
 ```
 npm run env:cli -- wp cloudinary analytics-events --clear
 ```
4. Perform the action in the browser.
5. Read what got captured:
 ```
 npm run env:cli -- wp cloudinary analytics-events
 ```
6. Eyeball the JSON line(s) — check `event_name`, `event_category`, and the event-specific params.

#### Example: Image Settings save

- In wp-admin, go to **Cloudinary → Image Settings**.
- Change something (e.g. Image Format → WebP) and click Save.
- Read the log — expect two lines:
  - `settings_saved` — `event_category: "settings"`, `page: "image"`, `changed_keys: ["image_format"]` (or whatever you changed).
  - `transformation_applied` — `scope: "global"` — this fires *additionally* when the changed key is one of the transformation-related fields (format/quality/etc.), per the diff check in `Admin::save_settings()`.

Same pattern works for any event: swap step 4 for the relevant UI action (or a direct REST call via the browser console using `window.cldData.analytics.nonce` / `.endpoint` if the UI path is awkward — see any of the `*-analytics.spec.js` files for exact REST routes per
event).


php/class-connect.php
```diff
@@ -591,6 +650,23 @@ public function check_status() {
 		$status = $this->test_ping();
 		$this->settings->get_setting( 'status' )->save_value( $status );
 
+		if ( is_wp_error( $status ) ) {
+			$analytics = $this->plugin->get_component( 'analytics' );
+			if ( $analytics ) {
+				$code = $status->get_error_code();
+				$analytics->track(
+					'connectivity_check_failed',
```
utkarshcloudinary, line 676: **`connectivity_check_failed` (ping) is not cron-only — it fires on every failed interactive connection test.**

`check_status()` is also called from `test_connection()` (php/class-connect.php:526), which runs on every `rest_test_connection` (wizard live-typing check) and on `verify_connection()` saves. A user typing a bad URL in the wizard emits both `connection_test_result` (error) *and* this event with `check_type: 'ping'`. The PR body/QA notes describe this event as cron-only.

Either gate the track on the cron context (e.g. `wp_doing_cron()` or only when invoked via the `cloudinary_status` hook), or accept and document the overlap — as-is it will inflate connectivity-failure metrics with interactive typos.

php/sync/class-sync-queue.php
```diff
@@ -556,6 +702,7 @@ public function stop_maybe( $type = 'queue' ) {
 	protected function shutdown_queue( $type = 'queue' ) {
 		if ( 'queue' === $type ) {
 			delete_option( self::$queue_enabled );
+			$this->track_run_completed();
```
utkarshcloudinary, line 744: **Stale run state corrupts the next `sync_completed`.**

This covers the queue-drained path, but the manual-stop path — `rest_start_sync()` with the queue disabled goes straight to `stop_queue()` (line ~718), not `shutdown_queue()` — never clears `RUN_STARTED_KEY`/`RUN_TALLY_KEY` and never emits.

The next run's `mark_run_started()` then no-ops (option already exists), so its `sync_completed` reports the **old** start time and accumulated tallies. Suggest clearing run state in `stop_queue( 'queue' )` alongside `delete_option( self::$queue_key )`, or explicitly in the manual-stop branch.

php/sync/class-push-sync.php
```diff
@@ -183,6 +183,24 @@ public function rest_start_sync( \WP_REST_Request $request ) {
 			$this->queue->stop_queue();
 		} else {
 			$state['success'] = $this->queue->start_queue( $type );
+
+			if ( $state['success'] ) {
+				$this->queue->mark_run_started( $type );
```
utkarshcloudinary: **Ordering: `mark_run_started()` runs after `start_queue()` has already spawned threads.**

`start_queue()` immediately kicks off background threads whose `log_sync_result()` → `tally_run_result()` calls no-op until `RUN_STARTED_KEY` exists — early results are silently dropped from the tally.

Set the run state before starting the queue (and clean it up on the `! $started` failure path so a failed start doesn't leave a phantom run behind).

php/sync/class-sync-queue.php
```diff
@@ -192,6 +211,94 @@ public function is_enabled() {
...
+	/**
+	 * Adds a sync result to the current run's success/error tally.
+	 *
+	 * A no-op if no run is currently being tracked.
+	 *
+	 * @param bool $success Whether the sync operation succeeded.
+	 *
+	 * @return void
+	 */
+	public function tally_run_result( $success ) {
```
utkarshcloudinary, line 255: **Tally read-modify-write races across sync threads.**

The queue runs multiple concurrent threads (separate requests); `get_option` → increment → `update_option` will lose updates under concurrency, undercounting `total_synced`/`total_errors`.

Either use an atomic `$wpdb` increment, or document the counts as approximate — acceptable as a known limitation if noted.

php/sync/class-sync-queue.php
```diff
@@ -200,6 +307,45 @@ public function is_enabled() {
...
+	 * Hooked to the generic per-save filter (fires for every changed setting),
+	 * filtered down to the sync-relevant slugs we care about.
+	 *
+	 * @param mixed   $new_value     The new value.
+	 * @param mixed   $current_value The current value.
+	 * @param Setting $setting       The setting object.
+	 *
+	 * @return mixed
+	 */
+	public function track_setting_changed( $new_value, $current_value, $setting ) {
```
utkarshcloudinary, line 350: **`sync_settings_changed` fires on no-op saves.**

This hooks `cloudinary_settings_save_setting`, which fires from `Settings::set_pending()` — and `set_pending()` is called *unconditionally* by `Connect::rest_save_wizard()` (php/class-connect.php:210-211 sets `auto_sync` every time) and by every `Setting::save_value()`. Unlike `Admin::save_settings()`, those callers don't diff first, and this tracker never compares `$new_value` to `$current_value`.

Every wizard completion will emit `sync_settings_changed` for `auto_sync` even when nothing changed. Add a `$new_value !== $current_value` guard (noting `$current_value` may be null when the caller doesn't pass it).

php/assets/class-rest-assets.php
```diff
@@ -280,6 +339,11 @@ public function rest_get_caches( $request ) {
 		$current_page = $page ? $page : 1;
 		$data         = $this->get_assets( $parent->ID, $search, $current_page );
 
+		$analytics = get_plugin_instance()->get_component( 'analytics' );
+		if ( $analytics ) {
+			$analytics->track( 'cache_items_viewed', 'cache', null, array( 'cache_point' => (string) $url ) );
```
utkarshcloudinary: **`cache_items_viewed` overcounts.** This fires on every `rest_get_caches()` call, including each pagination step and each search keystroke-triggered request. If the spec intends "user viewed a cache point", consider tracking only page 1 / no-search requests.

php/class-admin.php
```diff
@@ -138,6 +138,11 @@ public function rest_dismiss_notice( WP_REST_Request $request ) {
 		$duration = $request->get_param( 'duration' );
 
 		set_transient( $token, true, $duration );
+
+		$analytics = $this->plugin->get_component( 'analytics' );
+		if ( $analytics ) {
+			$analytics->track( 'notice_dismissed', 'settings', null, array( 'notice_id' => (string) $token ) );
```
utkarshcloudinary: Minor: `$token` is unsanitized user input going into the event (and into `set_transient()` — pre-existing). Cheap win: `sanitize_key( $token )` for `notice_id`.

src/js/deactivate.js
```diff
@@ -92,7 +93,18 @@ const Deactivate = {
...
+				if ( 'true' === context.modal.dataset.connected ) {
+					Analytics.track(
+						'deactivation_skipped',
+						{},
+						'deactivation'
+					);
+				}
+				// Give the fire-and-forget analytics request a moment to
+				// leave the browser before the page navigates away.
+				setTimeout( function () {
```
utkarshcloudinary: Minor: the 150 ms `setTimeout` is a race with navigation — slow admin-ajax/REST setup can still lose the event. `navigator.sendBeacon()` or `fetch( …, { keepalive: true } )` is the reliable pattern for exit events. Acceptable as fail-silent best-effort, but flaky by design.

php/class-deactivation.php
```diff
@@ -328,7 +328,19 @@ public function render_not_connected() {
...
+		$dependencies = array_filter(
+			$asset['dependencies'],
+			static function ( $dependency ) {
+				return false === strpos( $dependency, '/' );
+			}
+		);
+		// The deactivation script tracks analytics events, which requires the
+		// main `cloudinary` handle (and its `cldData` inline script) to be
+		// enqueued — it otherwise has no reason to load on the Plugins screen.
+		$dependencies[] = 'cloudinary';
```
utkarshcloudinary, line 341: Note: this pulls the full `cloudinary` bundle (plus jquery/wp-util) onto plugins.php for every site with the plugin installed, just to get `cldData` + the analytics bridge. Works, but a slim analytics-only handle would be lighter on a core screen. Fine to ship as-is if the size is acceptable.

php/class-admin.php
```diff
@@ -462,6 +495,80 @@ protected function save_settings( $submission, $data ) {
...
 	}
 
+	/**
+	 * Settings slugs (by page) that directly inject transformation qualifiers
+	 * (format, quality, freeform string) into delivered URLs, as opposed to
+	 * plain feature-enablement toggles.
+	 *
+	 * @var array
+	 */
+	const GLOBAL_TRANSFORMATION_SLUGS = array(
```
utkarshcloudinary: Nit: class constant declared mid-class between methods — move it up with the other constants at the top of the class.

utkarshcloudinary (approved):
Looks good! @gabrielcld2 

I reviewed with help of AI [which I reviewed] - left some comments.

php/class-connect.php
gabrielcld2: Addressed here [6d30260](https://github.com/cloudinary/cloudinary_wordpress/pull/1226/commits/6d30260712ae90c2b28fb686ce91591d0dd4551c). 
Change: Split the tracking out of `check_status()` into a new `cron_check_status()` wrapper, and point the cron registration at the wrapper instead.

php/sync/class-sync-queue.php
gabrielcld2: Addressed here [6d30260](https://github.com/cloudinary/cloudinary_wordpress/pull/1226/commits/6d30260712ae90c2b28fb686ce91591d0dd4551c).
Call `shutdown_queue('queue')` instead, which closes out the tracked run before stopping.

php/sync/class-push-sync.php
gabrielcld2: Addressed here [6d30260](https://github.com/cloudinary/cloudinary_wordpress/pull/1226/commits/6d30260712ae90c2b28fb686ce91591d0dd4551c).
Change: Reorder, set the run state first.

php/sync/class-sync-queue.php
gabrielcld2: Addressed here [6d30260](https://github.com/cloudinary/cloudinary_wordpress/pull/1226/commits/6d30260712ae90c2b28fb686ce91591d0dd4551c).
Change: Split into two scalar options, incremented atomically via raw SQL instead of a PHP round-trip.

php/sync/class-sync-queue.php
gabrielcld2: Fixed: Added a diff guard, with a fallback for callers that don't pass `$current_value`

php/assets/class-rest-assets.php
gabrielcld2: Fixed: Only track on page 1 with no search term

php/class-admin.php
gabrielcld2: Fixed!

src/js/deactivate.js
gabrielcld2: Fixed: Added `Analytics.trackReliable()`, which uses `navigator.sendBeacon()` instead of `fetch()`

php/class-deactivation.php
gabrielcld2: Leaving this as is for now. reworking the bundle might be a bit too much for this feature. Might be something worth doing someday though!

php/class-admin.php
gabrielcld2: Fixed Moved up next to `NOTICE_SLUG/SETTINGS_DATA` at the top of the class

php/sync/class-sync-queue.php
```diff
@@ -192,6 +224,94 @@ public function is_enabled() {
...
+		global $wpdb;
+
+		$wpdb->query( // phpcs:ignore WordPress.DB
+			$wpdb->prepare(
+				"INSERT INTO {$wpdb->options} (option_name, option_value, autoload) VALUES (%s, '1', 'no')
+				 ON DUPLICATE KEY UPDATE option_value = option_value + 1",
+				$option_name
+			)
+		);
+		wp_cache_delete( $option_name, 'options' );
```
utkarshcloudinary, line 279: **`notoptions` cache poisoning — `sync_completed` reports 0/0 on persistent-object-cache sites.**

`mark_run_started()` calls `delete_option()` on both tally keys, and core's `delete_option()` **adds the key to the `notoptions` cache**. This raw-SQL increment only clears the per-option cache — the key stays in `notoptions`, so on sites with a persistent object cache (Redis/Memcached), the later `get_option( RUN_TALLY_SYNCED_KEY )` in `track_run_completed()` short-circuits on `notoptions` and returns the default `0`. Every `sync_completed` would report `total_synced: 0, total_errors: 0`.

After the insert, also purge the key from `notoptions`, e.g.:
```php
$notoptions = wp_cache_get( 'notoptions', 'options' );
if ( is_array( $notoptions ) && isset( $notoptions[ $option_name ] ) ) {
	unset( $notoptions[ $option_name ] );
	wp_cache_set( 'notoptions', $notoptions, 'options' );
}
```
(wp-env/CI won't catch this — no persistent object cache there, per-request caches hide it.)

php/class-admin.php
```diff
@@ -134,10 +146,15 @@ public function rest_endpoints( $endpoints ) {
 	 * @param WP_REST_Request $request The request object.
 	 */
 	public function rest_dismiss_notice( WP_REST_Request $request ) {
-		$token    = $request->get_param( 'token' );
+		$token    = sanitize_key( $request->get_param( 'token' ) );
```
utkarshcloudinary: **`sanitize_key()` breaks notice-dismissal persistence.**

The dismiss token is the notice setting's slug (`class-notice.php:48` → `get_slug()`), which is a **dotted** path like `_cld_notices.0`. `sanitize_key()` strips dots, so the transient is written as `_cld_notices0` — but `Notice::is_enabled()` (php/ui/component/class-notice.php:135) checks `get_transient( $this->setting->get_slug() )` with the dot intact.

Result: dismissible notices with a real duration (e.g. the `MONTH_IN_SECONDS` usage-warning notices, php/class-connect.php:988) reappear on every page load after dismissal.

Use a dot-preserving sanitizer instead, e.g. `preg_replace( '/[^a-z0-9_.\-]/i', '', (string) $token )`; or sanitize only the `notice_id` event param and keep the raw token for the transient key.

utkarshcloudinary (approved):
Looks good!

php/sync/class-sync-queue.php
gabrielcld2: Good catch! Should be fixed now [9eb38fa](https://github.com/cloudinary/cloudinary_wordpress/pull/1226/commits/9eb38fa5e7ea87154801eab22ba7a62b519d1929)

php/class-admin.php
gabrielcld2: fixed here [9eb38fa](https://github.com/cloudinary/cloudinary_wordpress/pull/1226/commits/9eb38fa5e7ea87154801eab22ba7a62b519d1929)

PR #1251, opened 2026-08-25
Add WordPress unit testing with PHPUnit via wp-env

utkarshcloudinary:
## Approach

Adds a PHPUnit harness that runs against the WordPress core test suite that wp-env already mounts into its containers, so local runs and CI use the same environment as the existing e2e suite.

wp-env ships PHPUnit 10 globally, but the WP 7.0 core test suite calls `PHPUnit\Util\Test::parseTestMethodAnnotations()`, which PHPUnit 10 removed. That call runs from `expectDeprecated()` on every `set_up()`, so every test would fatal on the global binary. The plugin therefore pins its own PHPUnit 9.6 plus the Yoast polyfills as dev dependencies and runs `vendor/bin/phpunit` inside the container.

- `phpunit.xml.dist` and `tests/phpunit/bootstrap.php`, which reads the `WP_TESTS_DIR` that wp-env exports and loads the plugin on `muplugins_loaded`
- `phpunit/phpunit ^9.6@stable` and `yoast/phpunit-polyfills ^1.1@stable` in `require-dev`. The `@stable` suffix is needed because the repo sets `minimum-stability: dev`, which otherwise resolves to dev branch builds
- 18 tests covering the `Utils` helpers, Cloudinary CDN URL generation, transformation strings, CNAME handling, image tag attribute rebuilding, and the `cloudinary_bypass_seo_url` filter
- New `unit` CI job, with a `~/.wp-env` cache so runs do not re-clone roughly 1 GB of WordPress and core PHPUnit sources
- `tests/phpunit/` excluded from PHPCS, matching the existing treatment of the PHPStan stubs

The tests need no Cloudinary account. The URL build path reads only `cloud_name`, `cname` and `private_cdn`, never the API key or secret, so a stub connection is enough and the CI job needs no secrets. That also means it works on forked PRs. Real account coverage stays in the Playwright suite.

## QA notes

Run the suite locally:

```
npm install
npm run env:start
npm run test:unit
```

Expected: `OK (18 tests, 35 assertions)`.

To confirm a failure stops the build, add a failing assertion to any test in `tests/phpunit/tests/` and re-run. The command should exit non-zero, which fails the `Unit (PHPUnit)` CI job. This was verified during development: a deliberate failure gave exit code 1 and `Tests: 19, Failures: 1`.

Also worth checking that the `unit` job appears and passes on this PR.

## Follow-ups, not addressed here

- If branch protection lists required checks by name, `unit` needs adding to that list, otherwise a red job will not block merges.
- Pre-existing and unrelated to this change: `composer lint` runs out of memory after a local `composer phpstan` run, because PHPCS scans the generated `.phpstan-cache/` directory. It is gitignored but not in the PHPCS exclusions. CI is unaffected since the jobs are separate.
- Pre-existing: `composer audit` flags CVE-2026-45293 (high) against `wp-coding-standards/wpcs`, fixed in 3.4.1.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,6 +80,54 @@ jobs:
       - name: Run PHPStan
         run: composer phpstan
 
+  unit:
+    name: Unit (PHPUnit)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+          cache: 'npm'
+
+      - name: Cache Composer
+        uses: actions/cache@v4
+        with:
+          path: ~/.composer/cache
+          key: ${{ runner.os }}-composer-unit-${{ hashFiles('**/composer.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-composer-unit-
+
+      # wp-env downloads WordPress and the core PHPUnit suite into this
+      # directory, which is roughly a gigabyte of git clones. Caching it keeps
+      # `env:start` from re-cloning them on every run. The key covers the
+      # wp-env config and version, because a change to either means different
+      # sources have to be downloaded.
+      - name: Cache wp-env sources
+        uses: actions/cache@v4
+        with:
+          path: ~/.wp-env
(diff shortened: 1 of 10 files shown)


gabrielcld2 (approved):
👍 Everything works as expected

PR #1252, opened 2026-08-26
Fix: don't overwrite an unrelated Cloudinary asset on public ID collision

### LINKED ISSUE #1241: Cloudinary asset being replaced
## Bug Description

PR [#1182](https://github.com/cloudinary/cloudinary_wordpress/pull/1182) causes Cloudinary assets to be overwritten when a new WordPress upload reuses the same filename.

WordPress only uniquifies filenames within the same month. If `test.jpg` is uploaded in April, Cloudinary creates public ID `test`. A second `test.jpg` in the same month is renamed to `test-1.jpg` (public ID `test-1`). A later upload of `test.jpg` in May keeps the original filename, so Cloudinary again gets public ID `test`.

After #1182, the plugin treats an existing public ID as a replacement and overwrites the original Cloudinary asset instead of creating a distinct one.

## Expected Behaviour

Uploading a new file with the same basename in a different month should not overwrite an existing Cloudinary asset. The plugin should generate a unique public ID (or otherwise treat it as a new asset), even when WordPress reuses the original filename.

## Steps to reproduce

1. Upload `test.jpg` to the WordPress Media Library in month A (e.g. April). Confirm it syncs to Cloudinary with public ID `test`.
2. Upload a different image also named `test.jpg` in the same month. Confirm WordPress renames it to `test-1.jpg` and Cloudinary uses public ID `test-1`.
3. Upload another different image named `test.jpg` in a later month (e.g. May). Confirm WordPress keeps the filename `test.jpg`.
4. Check the original Cloudinary asset with public ID `test`.
5. See that it has been overwritten by the May upload instead of remaining the April image.

## Additional context

This is a regression from https://github.com/cloudinary/cloudinary_wordpress/pull/1182. The change detects an existing public ID and assumes a replacement, which is incorrect when WordPress reuses a filename across months.

 - WordPress version:
 - Plugin version:
 - Gutenberg plugin version (if applicable):
 - PHP version:
 - Plugin settings:

gabrielcld2:
## Summary

- [PR #1182](https://github.com/cloudinary/cloudinary_wordpress/pull/1182) fixed a runaway-duplicate bug (an attachment retrying an interrupted upload kept creating suffixed Cloudinary duplicates) by overwriting the conflicting asset whenever the attachment had no locally-saved `_public_id`.
- That signal isn't reliable on its own: **every never-synced attachment** also has no `_public_id`, not just one recovering from a crashed upload. So a brand new attachment that happens to derive the same public ID as an older, unrelated asset (e.g. WordPress reuses a filename across different months) also takes the overwrite path and silently destroys that older asset.
- This adds a check, `Upload_Sync::is_matching_existing_asset()`, that only takes the overwrite path when the conflicting Cloudinary asset is confirmed to be the local file. Same file, lost record (PR #1182's case) → confirmed match → still safely overwritten. Different file, coincidental public ID collision (this issue) → no match → falls back to the pre-#1182 suffix behavior instead, so the older asset is preserved.

Fixes #1241

## Review follow-up

Addressed in a follow-up commit after review:

- **Blocker:** the check now resolves the local file the same way `Api::upload()` does (`Media::get_upload_file_path()`, shared by both), instead of always using `get_attached_file()`. That matters for any image over `big_image_size_threshold` (2560px default) — WordPress attaches the `-scaled` copy but Cloudinary is sent the unscaled original, so comparing against the attached file broke the #1182 recovery path for large images.
- Confirms with the response's `etag` (MD5 of the stored asset) once byte sizes already match, closing the remaining false-positive where two unrelated files coincidentally share a byte count.
- Logs via `Utils::log()` when the check bails out for a missing `bytes` field, so a future API response change surfaces in the debug log instead of silently reviving the duplicate-per-cycle bug.
- Docblock narrowed and marked `@internal`.

Addressed in a second follow-up commit after a further review pass:

- **vip:// paths:** `get_upload_file_path()` can return a `vip://` stream-wrapper path on WordPress VIP. Hashing it with `md5_file()` pulls the whole object over the network, and a failed read returns `false` rather than throwing — which the etag check would misread as a content mismatch, reintroducing the #1182 duplicate-per-cycle bug on VIP specifically. Now skips the hash and relies on byte size alone for `vip://` paths.
- **Return type:** `Media::get_upload_file_path()`'s docblock said `@return string`; corrected to `string|false` to match the underlying `get_attached_file()`/`wp_get_original_image_path()` core functions.
- **Content equality isn't ownership:** two unrelated attachments can hold byte-identical files that derive the same public ID — both would pass the byte/etag check and both would overwrite the same asset, clobbering each other's context and version. Added `is_solely_linked_to()`, mirroring the ownership guard `Delete_Sync::delete_asset()` already uses (`get_linked_attachments()`), so the overwrite path also requires that no *other* attachment is already tracked as linked to that public ID.

### Scenario 1 — the reported bug (#1241) is fixed

1. Upload image A named `test.jpg`. Wait for sync. Note its ID as `<A_ID>`.
2. Upload a second, different image (any filename). Note its ID as `<B_ID>`. Then force it into "first-ever upload of `test.jpg`" state:
   ```bash
   npx wp-env run cli wp eval '
   $id      = <B_ID>;
   $old     = get_attached_file( $id );
   $new_dir = dirname( $old ) . "-b";
   wp_mkdir_p( $new_dir );
   $new = trailingslashit( $new_dir ) . "test.jpg";
   rename( $old, $new );
   update_attached_file( $id, $new );
   $m = get_post_meta( $id, "_cloudinary", true );
   unset( $m["_public_id"], $m["_sync_signature"] );
   update_post_meta( $id, "_cloudinary", $m );
   '
   ```
3. Trigger B's sync:
   ```bash
   npx wp-env run cli wp eval 'Cloudinary\get_plugin_instance()->get_component("sync")->managers["push"]->process_assets(<B_ID>);'
   ```
4. Compare public IDs:
   ```bash
   npx wp-env run cli wp eval 'echo get_post_meta(<A_ID>, "_cloudinary", true)["_public_id"] . "\n";'
   npx wp-env run cli wp eval 'echo get_post_meta(<B_ID>, "_cloudinary", true)["_public_id"] . "\n";'
   ```
   - **On `develop` (bug present):** both print `test` — B silently claimed A's public ID and overwrote its Cloudinary asset.
   - **On this branch (fixed):** A stays `test`, B gets a suffixed public ID (e.g. `test_<id><random>`) — both assets survive independently.

### Scenario 2 — PR #1182's original crash-recovery fix still works

Use an image **larger than 2560px** on its longest side for step 1 — that's what exercises the `-scaled` file path the review feedback flagged, and what scenario 2 originally missed.

1. Upload the large image. Wait for sync. Note its ID as `<X_ID>`.
2. Simulate a lost record (same file, nothing changes — this is the "meta lost" case, not the collision case):
   ```bash
   npx wp-env run cli wp eval '
   $id = <X_ID>;
   $m = get_post_meta( $id, "_cloudinary", true );
   unset( $m["_public_id"], $m["_sync_signature"] );
   update_post_meta( $id, "_cloudinary", $m );
   '
   ```
3. Re-trigger the sync:
   ```bash
   npx wp-env run cli wp eval 'Cloudinary\get_plugin_instance()->get_component("sync")->managers["push"]->process_assets(<X_ID>);'
   ```
4. Check the saved public_id:
   ```bash
   npx wp-env run cli wp eval 'echo get_post_meta(<X_ID>, "_cloudinary", true)["_public_id"] . "\n";'
   ```
   - **Expected (fix intact):** no suffix — the attachment recovered by overwriting its own orphaned asset, same as before.
   - A suffixed public_id here would mean the crash-recovery path regressed.

Both scenarios were run and confirmed against a real Cloudinary account, including scenario 2 re-run with a 3200×1800px/1.74MB image (large enough to trigger WordPress's `-scaled` file, per the review feedback): the recovered public_id came back unsuffixed, confirming the crash-recovery path now also works correctly for scaled images.

### Automated tests

`tests/phpunit/tests/test-upload-sync.php` covers `is_matching_existing_asset()`: matching/mismatched bytes, a missing `bytes`/`public_id` field, a missing local file, matching/mismatched `etag`, a scaled-image case (compares against the unscaled original, not the attached `-scaled` file), an ownership-conflict case (another attachment already linked to the public ID, even with byte-identical content), and a `vip://` path case (matches on bytes alone, ignoring a deliberately wrong etag, via a fake stream wrapper). 29 tests. Run with `npm run env:start && npm run test:unit`.

### DIFF
--- a/php/class-media.php
+++ b/php/class-media.php
@@ -496,6 +496,27 @@ class Media extends Settings_Component implements Setup {
 	}
 
 	/**
+	 * Get the local file path used to upload an attachment.
+	 *
+	 * Mirrors the file resolution in Connect\Api::upload(): the unscaled original when
+	 * `cloudinary_use_original_image` allows it, the attached file otherwise -- e.g. the
+	 * `-scaled` copy WordPress creates for images over `big_image_size_threshold`.
+	 *
+	 * @param int $attachment_id The attachment ID.
+	 *
+	 * @return string
+	 */
+	public function get_upload_file_path( $attachment_id ) {
+		/** This filter is documented in php/connect/class-api.php */
+		$use_original = apply_filters( 'cloudinary_use_original_image', true, $attachment_id );
+		if ( $use_original && function_exists( 'wp_get_original_image_path' ) && wp_attachment_is_image( $attachment_id ) ) {
+			return wp_get_original_image_path( $attachment_id );
+		}
+
+		return get_attached_file( $attachment_id );
+	}
+
+	/**
 	 * Get the Cloudinary delivery type.
 	 *
 	 * @param int $attachment_id The attachment ID.
--- a/php/class-media.php
+++ b/php/class-media.php
@@ -504,7 +504,7 @@ class Media extends Settings_Component implements Setup {
 	 *
 	 * @param int $attachment_id The attachment ID.
 	 *
-	 * @return string
+	 * @return string|false
 	 */
 	public function get_upload_file_path( $attachment_id ) {
 		/** This filter is documented in php/connect/class-api.php */
--- a/php/sync/class-upload-sync.php
+++ b/php/sync/class-upload-sync.php
@@ -391,8 +391,12 @@ class Upload_Sync {
 	 * Check whether a Cloudinary "existing" asset is likely this attachment's own local file.
 	 *
 	 * Used to tell apart an orphan left by this same attachment's previously interrupted upload
-	 * (safe to overwrite) from an unrelated asset that happens to share the same derived public
-	 * ID, e.g. WordPress reusing a filename across months (must not be overwritten).
+	 * of the default (non "folder"/"cloud_name") sync type (safe to overwrite) from an unrelated
+	 * asset that happens to share the same derived public ID, e.g. WordPress reusing a filename
+	 * across months (must not be overwritten). Only called once a public_id is unrecorded, so in
+	 * practice this only ever runs for that default sync type; the other types always have one.
+	 *
+	 * @internal Reachable for testing; not intended to be called from outside this class.
 	 *
 	 * @param int   $attachment_id The attachment ID.
 	 * @param array $result        The Cloudinary upload result.
@@ -401,14 +405,23 @@ class Upload_Sync {
 	 */
 	public function is_matching_existing_asset( $attachment_id, $result ) {
 		if ( empty( $result['bytes'] ) ) {
+			Utils::log(
+				sprintf( 'Cloudinary upload result for attachment %d has no "bytes" field; treating as a non-matching asset.', $attachment_id ),
+				'upload-sync-existing-asset-check'
+			);
+
 			return false;
 		}
-		$file = get_attached_file( $attachment_id );
+		$file = $this->media->get_upload_file_path( $attachment_id );
 		if ( empty( $file ) || ! file_exists( $file ) ) {
 			return false;
 		}
+		if ( (int) filesize( $file ) !== (int) $result['bytes'] ) {
+			return false;
+		}
 
-		return (int) filesize( $file ) === (int) $result['bytes'];
+		// Bytes alone can coincide between unrelated files; confirm with the content hash when available.
+		return empty( $result['etag'] ) || md5_file( $file ) === $result['etag'];
 	}
 
 	/**
--- a/php/sync/class-upload-sync.php
+++ b/php/sync/class-upload-sync.php
@@ -412,6 +412,12 @@ class Upload_Sync {
 
 			return false;
 		}
+		// Byte-identical content between two unrelated attachments isn't proof of ownership: the
+		// second overwrite would still clobber the first's context and advance its version. Only
+		// proceed if no other attachment already claims this public ID.
+		if ( ! $this->is_solely_linked_to( $attachment_id, empty( $result['public_id'] ) ? null : $result['public_id'] ) ) {
+			return false;
+		}
 		$file = $this->media->get_upload_file_path( $attachment_id );
 		if ( empty( $file ) || ! file_exists( $file ) ) {
 			return false;
@@ -419,9 +425,42 @@ class Upload_Sync {
 		if ( (int) filesize( $file ) !== (int) $result['bytes'] ) {
 			return false;
 		}
+		// Hashing a vip:// stream wrapper path pulls the whole object over the network; a failed
+		// read returns false rather than throwing, which would wrongly read as a mismatch. Bytes
+		// alone is the safer signal to rely on there.
+		if ( false !== strpos( $file, 'vip://' ) ) {
+			return true;
+		}
 
 		// Bytes alone can coincide between unrelated files; confirm with the content hash when available.
 		return empty( $result['etag'] ) || md5_file( $file ) === $result['etag'];
+	}
+
+	/**
+	 * Check that no other attachment is already tracked as linked to a public ID.
+	 *
+	 * Mirrors the ownership guard Delete_Sync::delete_asset() uses before destroying an asset.
+	 *
+	 * @param int         $attachment_id The attachment ID.
+	 * @param string|null $public_id     The public ID to check.
+	 *
+	 * @return bool
+	 */
+	protected function is_solely_linked_to( $attachment_id, $public_id ) {
+		if ( empty( $public_id ) ) {
+			return false;
+		}
+		$linked = $this->media->get_linked_attachments( $public_id );
+		if ( count( $linked ) > 1 ) {
+			// More than one attachment already shares this public ID.
+			return false;
+		}
+		if ( 1 === count( $linked ) && (int) $attachment_id !== (int) $linked[0] ) {
+			// Exactly one other attachment is already linked to it.
+			return false;
+		}
+
+		return true;
 	}
 
 	/**
--- a/php/sync/class-upload-sync.php
+++ b/php/sync/class-upload-sync.php
@@ -338,9 +338,14 @@ class Upload_Sync {
 
 			// Check that this wasn't an existing.
 			if ( ! empty( $result['existing'] ) ) {
-				// If no public_id is recorded in WordPress, this asset in Cloudinary is from a
-				// failed previous upload. Overwrite it instead of creating a suffixed duplicate.
-				if ( empty( $suffix ) && ! $this->media->get_post_meta( $attachment_id, Sync::META_KEYS['public_id'], true ) ) {
+				// A missing public_id in WordPress isn't enough on its own to prove the conflicting
+				// Cloudinary asset is an orphan of this attachment's own failed upload -- any never
+				// synced attachment also has no public_id. Only treat it as our own orphan, safe to
+				// overwrite, when the existing asset's file size also matches the local file.
+				if ( empty( $suffix )
+					&& ! $this->media->get_post_meta( $attachment_id, Sync::META_KEYS['public_id'], true )
+					&& $this->is_matching_existing_asset( $attachment_id, $result )
+				) {
 					return $this->upload_asset( $attachment_id, $type, null, true );
 				}
 				// Add a suffix and try again.
@@ -380,6 +385,30 @@ class Upload_Sync {
 		}
 
 		return $result;
+	}
+
+	/**
+	 * Check whether a Cloudinary "existing" asset is likely this attachment's own local file.
+	 *
+	 * Used to tell apart an orphan left by this same attachment's previously interrupted upload
+	 * (safe to overwrite) from an unrelated asset that happens to share the same derived public
+	 * ID, e.g. WordPress reusing a filename across months (must not be overwritten).
+	 *
+	 * @param int   $attachment_id The attachment ID.
+	 * @param array $result        The Cloudinary upload result.
+	 *
+	 * @return bool
+	 */
+	public function is_matching_existing_asset( $attachment_id, $result ) {
+		if ( empty( $result['bytes'] ) ) {
+			return false;
+		}
+		$file = get_attached_file( $attachment_id );
+		if ( empty( $file ) || ! file_exists( $file ) ) {
+			return false;
+		}
+
+		return (int) filesize( $file ) === (int) $result['bytes'];
 	}
 
 	/**
--- a/tests/phpunit/tests/test-upload-sync.php
+++ b/tests/phpunit/tests/test-upload-sync.php
@@ -6,7 +6,8 @@
  * Cloudinary asset blocking an upload (existing: true) is safe to overwrite. It should only
  * be treated as this attachment's own orphaned upload -- not an unrelated asset that happens
  * to share the same derived public ID, e.g. WordPress reusing a filename across months (see
- * GitHub issue #1241).
+ * GitHub issue #1241), and not an unrelated attachment that happens to hold byte-identical
+ * content and already owns that public ID.
  *
  * The rest of upload_asset() talks to the Cloudinary API over HTTP and is covered by the
  * Playwright suite in tests/e2e.
@@ -66,6 +67,19 @@ class Test_Upload_Sync extends WP_UnitTestCase {
 	}
 
 	/**
+	 * Mark an attachment as linked to a public ID, the way a completed upload_asset() call
+	 * would via its trackable postmeta key -- what get_linked_attachments() looks up.
+	 *
+	 * @param int    $attachment_id The attachment ID.
+	 * @param string $public_id     The public ID.
+	 *
+	 * @return void
+	 */
+	protected function link_attachment_to_public_id( $attachment_id, $public_id ) {
+		update_post_meta( $attachment_id, '_' . md5( $public_id ), true );
+	}
+
+	/**
 	 * An existing asset whose byte size matches the local file is treated as this attachment's
 	 * own orphaned upload, so it's safe to overwrite. No etag in the result falls back to the
 	 * byte comparison alone.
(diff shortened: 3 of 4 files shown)


php/sync/class-upload-sync.php
```diff
@@ -382,6 +387,30 @@ function ( $is_synced, $post_id ) use ( $attachment_id ) {
...
+	 * @param int   $attachment_id The attachment ID.
+	 * @param array $result        The Cloudinary upload result.
+	 *
+	 * @return bool
+	 */
+	public function is_matching_existing_asset( $attachment_id, $result ) {
+		if ( empty( $result['bytes'] ) ) {
+			return false;
+		}
+		$file = get_attached_file( $attachment_id );
```
utkarshcloudinary: **Blocker:** this does not resolve the same file that the upload actually sent, so the #1182 recovery path stops working for large images.

`Api::upload()` applies the `cloudinary_use_original_image` filter (default `true`) and then uses `wp_get_original_image_path()` for images:

```php
// php/connect/class-api.php:529-534, 564-565
$use_original = apply_filters( 'cloudinary_use_original_image', true, $attachment_id );
...
$get_path_func = $use_original && function_exists( 'wp_get_original_image_path' ) ? 'wp_get_original_image_path' : 'get_attached_file';
```

Core diverges the two whenever `original_image` is set in the attachment metadata: `wp_get_original_image_path()` returns the untouched original, `get_attached_file()` returns the `-scaled` file.

So for any image above `big_image_size_threshold` (2560px default, which covers most phone and camera uploads) this compares the `-scaled` bytes against the original's bytes. They never match, `is_matching_existing_asset()` returns `false`, and the attachment falls back to the suffix branch. That is exactly the runaway-duplicate behaviour PR #1182 existed to stop.

The QA notes in the description do not catch this because both scenarios use a small image that never gets scaled.

Mirroring the upload's own resolution fixes it:

```php
$use_original = apply_filters( 'cloudinary_use_original_image', true, $attachment_id );
if ( $use_original && function_exists( 'wp_get_original_image_path' ) && wp_attachment_is_image( $attachment_id ) ) {
	$file = wp_get_original_image_path( $attachment_id );
} else {
	$file = get_attached_file( $attachment_id );
}
```

`Media::get_attachment_file_size()` (php/class-media.php:475-496) already contains this branch. Worth extracting it into one shared helper rather than duplicating a third copy. Note it caches into `_file_size` meta, so calling that method directly would feed a possibly stale value into a destructive decision.

Please also re-run QA scenario 2 with an image over 2560px, that is the case that currently fails.

php/sync/class-upload-sync.php
```diff
@@ -382,6 +387,30 @@ function ( $is_synced, $post_id ) use ( $attachment_id ) {
...
+		if ( empty( $file ) || ! file_exists( $file ) ) {
+			return false;
+		}
+
+		return (int) filesize( $file ) === (int) $result['bytes'];
```
utkarshcloudinary: **Consider:** byte size alone is a weak identity signal, in both directions.

- False positive: two unrelated files that derive the same public ID *and* happen to have the same byte count still take the overwrite path, and the older asset is still destroyed. That is the exact failure mode this PR exists to prevent, narrowed rather than closed.
- False negative: if the product environment applies an incoming transformation, or a `cloudinary_upload_options` filter adds `transformation`, the stored asset's bytes will not equal the local file's bytes, and the #1182 recovery silently stops working for that environment.

Cloudinary returns `etag` in the same response, which is the MD5 of the stored bytes. Gating on bytes first and then confirming with `md5_file()` closes the false positive completely, and the hash only runs on the collision path, never on a normal upload:

```php
if ( (int) filesize( $file ) !== (int) $result['bytes'] ) {
	return false;
}

return empty( $result['etag'] ) || md5_file( $file ) === $result['etag'];
```

php/sync/class-upload-sync.php
```diff
@@ -382,6 +387,30 @@ function ( $is_synced, $post_id ) use ( $attachment_id ) {
...
+	 * (safe to overwrite) from an unrelated asset that happens to share the same derived public
+	 * ID, e.g. WordPress reusing a filename across months (must not be overwritten).
+	 *
+	 * @param int   $attachment_id The attachment ID.
+	 * @param array $result        The Cloudinary upload result.
+	 *
+	 * @return bool
+	 */
+	public function is_matching_existing_asset( $attachment_id, $result ) {
+		if ( empty( $result['bytes'] ) ) {
```
utkarshcloudinary, line 407: **Nit:** the fallback direction is right, safe over destructive. But it makes the #1182 fix depend on an undocumented field of the `existing` response. If that field is ever trimmed, the recovery path dies silently, with no error and no log, and suffixed duplicates quietly come back.

A sync note or debug trace when the check bails out on a missing field would save the next person from bisecting two PRs to work out why.

php/sync/class-upload-sync.php
```diff
@@ -382,6 +387,30 @@ function ( $is_synced, $post_id ) use ( $attachment_id ) {
...
+	 * Used to tell apart an orphan left by this same attachment's previously interrupted upload
+	 * (safe to overwrite) from an unrelated asset that happens to share the same derived public
+	 * ID, e.g. WordPress reusing a filename across months (must not be overwritten).
+	 *
+	 * @param int   $attachment_id The attachment ID.
+	 * @param array $result        The Cloudinary upload result.
+	 *
+	 * @return bool
+	 */
+	public function is_matching_existing_asset( $attachment_id, $result ) {
```
utkarshcloudinary, line 406: **Nit:** `public` for a method with a single internal caller. Reasonable trade to make it reachable from the test, but it is now plugin API surface that has to keep its signature. An `@internal` note in the docblock would set the expectation.

Also worth narrowing the docblock's claim of generality: the `folder` and `cloud_name` sync types route to `Api::copy()`, which uploads from a Cloudinary URL, not a local file, and under `offload=cld` there may be no local file at all. In practice those types only run once a `public_id` is recorded, so the second clause of the condition short-circuits first and this is never reached, but the docblock reads as though it applies to any upload.

utkarshcloudinary (approved):
Review with help of AI and I validated those comments:

php/sync/class-upload-sync.php
gabrielcld2: Fixed in 9187312e — thanks for catching this. `is_matching_existing_asset()` now calls a new shared `Media::get_upload_file_path()`, which is the exact same resolution `Api::upload()` uses (`cloudinary_use_original_image` filter + `wp_get_original_image_path()` for images, falling back to `get_attached_file()`), so both now compare against the same file `Api::upload()` actually sent. `Api::upload()`'s own inline branch was replaced with a call to the same helper, so there's one canonical implementation instead of a third copy. Added a test (`test_matches_using_the_unscaled_original_for_a_scaled_image`) that fakes a scaled/original pair via `_wp_attachment_metadata['original_image']` and asserts the check matches on the original's bytes, not the attached (scaled) file's.\n\nRe-running QA scenario 2 with an image over 2560px on this commit — will report back once confirmed.

php/sync/class-upload-sync.php
gabrielcld2: Done in 9187312e — once bytes match, it now confirms with `md5_file()` against the response's `etag`, only computed on the collision path as you suggested. Kept it opt-in (`empty( $result['etag'] ) ||`) rather than required, so a response missing that field still falls back to the byte check instead of always failing closed — the false-negative you flagged (transformation-altered stored bytes) already broke the byte check on its own, so treating a missing etag as inconclusive-in-the-wrong-direction there wouldn't help; it would just add a second silent failure mode. Added two tests covering the etag match/mismatch cases.

php/sync/class-upload-sync.php
gabrielcld2: Added — 9187312e logs via `Utils::log()` (the existing debug-report mechanism, see `class-delivery.php`/`class-responsive-breakpoints.php` for the same pattern) when the check bails out for a missing `bytes` field, so a future response shape change shows up in the debug log instead of silently reverting to the old duplicate-per-cycle behavior.

php/sync/class-upload-sync.php
gabrielcld2: Both addressed in 9187312e: added an `@internal` note, and narrowed the docblock to note this only actually runs for the default sync type in practice, since `folder`/`cloud_name` route through `Api::copy()` and, as you noted, only ever get here once a public_id already exists (short-circuiting the first clause). Left it `public` per your call.

php/sync/class-upload-sync.php
gabrielcld2: Confirmed — re-ran QA scenario 2 with a 3200×1800px, 1.74MB image (large enough to get WordPress's `-scaled` treatment). The recovered public_id came back unsuffixed (`scaled-image-test`), so the crash-recovery path now works correctly for scaled images too. Updated the PR description accordingly.

php/sync/class-upload-sync.php
```diff
@@ -382,6 +387,43 @@ function ( $is_synced, $post_id ) use ( $attachment_id ) {
...
+		$file = $this->media->get_upload_file_path( $attachment_id );
+		if ( empty( $file ) || ! file_exists( $file ) ) {
+			return false;
+		}
+		if ( (int) filesize( $file ) !== (int) $result['bytes'] ) {
+			return false;
+		}
+
+		// Bytes alone can coincide between unrelated files; confirm with the content hash when available.
+		return empty( $result['etag'] ) || md5_file( $file ) === $result['etag'];
```
utkarshcloudinary: **warning:** content equality is not ownership.

Two unrelated attachments holding byte-identical files that derive the same public ID both pass this check. Example: the same stock image uploaded twice under the same filename in different months. Both get `overwrite => true` against the same Cloudinary asset.

The delivered bytes are identical, so this is not the data-loss case #1241 reports. The damage is narrower but real:

- `Media::get_upload_options()` sends `context` on every upload (`class-media.php:2881`). For a folder-synced attachment that context carries `caption`, `alt`, and `guid` (`get_context_options()`, `class-media.php:2804-2808`). The second attachment overwrites the first attachment's caption/alt/guid on the shared asset.
- The `version` advances, so the first attachment's stored `_version` meta goes stale and its cached URLs point at a superseded version.

Not asking for a redesign. Either scope the docblock to say this confirms the file content is the same, and that the remaining collision is accepted because the bytes are identical, or check `Media::get_linked_attachments( $public_id )` before taking the overwrite path, the same guard `Delete_Sync::delete_asset()` already uses at `class-delete-sync.php:95-103`.

php/sync/class-upload-sync.php
```diff
@@ -382,6 +387,43 @@ function ( $is_synced, $post_id ) use ( $attachment_id ) {
...
+	public function is_matching_existing_asset( $attachment_id, $result ) {
+		if ( empty( $result['bytes'] ) ) {
+			Utils::log(
+				sprintf( 'Cloudinary upload result for attachment %d has no "bytes" field; treating as a non-matching asset.', $attachment_id ),
+				'upload-sync-existing-asset-check'
+			);
+
+			return false;
+		}
+		$file = $this->media->get_upload_file_path( $attachment_id );
```
utkarshcloudinary: **warning:** `vip://` paths are not handled here, but `Api::upload()` handles them.

`get_upload_file_path()` can return a `vip://` path. `Api::upload()` detects that and calls `create_local_copy()` before touching the file (`class-api.php:572-578`). This check calls `file_exists()`, `filesize()`, and `md5_file()` straight on the stream wrapper path.

On VIP that means `md5_file()` pulls the whole object over the network on the collision path. If the wrapper cannot serve the read, `md5_file()` returns `false`, the comparison at line 424 fails, and the attachment silently falls back to the suffix path. That is the exact #1182 crash-recovery regression this PR is meant to preserve, reappearing only on VIP where it will be hardest to reproduce.

Suggest skipping the hash and returning on bytes alone when `strpos( $file, 'vip://' ) !== false`.

php/class-media.php
```diff
@@ -495,6 +495,27 @@ function_exists( 'wp_get_original_image_path' )
...
+	/**
+	 * Get the local file path used to upload an attachment.
+	 *
+	 * Mirrors the file resolution in Connect\Api::upload(): the unscaled original when
+	 * `cloudinary_use_original_image` allows it, the attached file otherwise -- e.g. the
+	 * `-scaled` copy WordPress creates for images over `big_image_size_threshold`.
+	 *
+	 * @param int $attachment_id The attachment ID.
+	 *
+	 * @return string
```
utkarshcloudinary: **warning:** return type is wrong.

Both `wp_get_original_image_path()` and `get_attached_file()` return `string|false`. `@return string` is inaccurate.

This is not cosmetic. `is_matching_existing_asset()` guards with `empty( $file )` so it is safe, but `Api::upload()` at `class-api.php:566` and `572` passes the result straight into `file_exists()` and `strpos()`. A caller trusting this docblock will skip the falsy guard. Use `@return string|false`.

tests/phpunit/tests/test-upload-sync.php
```diff
@@ -0,0 +1,195 @@
...
+ *
+ * @package Cloudinary
+ */
+
+use Cloudinary\Sync\Upload_Sync;
+
+/**
+ * Covers the existing-asset ownership check in Upload_Sync.
+ */
+class Test_Upload_Sync extends WP_UnitTestCase {
```
utkarshcloudinary: **warning:** the behaviour change itself has no test.

All seven tests target `is_matching_existing_asset()` in isolation. The change that fixes #1241 is the added condition in `upload_asset()` at `class-upload-sync.php:345-348`, and nothing here covers it. Deleting `&& $this->is_matching_existing_asset(...)` from that condition leaves this whole suite green while restoring the bug.

The API call is the obstacle, but not an insurmountable one: `$this->connect->api` is reachable, so a stubbed `api` returning `array( 'existing' => true, 'bytes' => ... )` on the first call and a normal result on the second would let you assert that a mismatched-bytes result produces a suffixed `public_id` and a matching one does not. That is exactly what the QA notes in the description verify by hand.

If that is out of scope for this PR, say so in the description so the gap is a known one rather than an assumed pass.

PR #1269, opened 2026-09-04
perf(e2e): Run specs in parallel workers and overlap CI setup

utkarshcloudinary:
Improves e2e test runtime in CI (no linked issue).

## Approach

- The e2e job took 5.5 to 6.5 min: ~2.5 min for the Playwright run with a single worker, ~2 min for setup run strictly in sequence. This PR makes the analytics specs worker-safe so most of the suite runs in parallel, and overlaps the independent setup steps.
- Per-worker analytics capture log. The mu-plugin reads a `cld_e2e_worker` cookie (browser and REST traffic) or a `CLD_E2E_WORKER` env var (WP-CLI via `docker exec`) and appends to `analytics-capture-<worker>.log`. Unmarked requests keep using the shared log. A cookie rather than `extraHTTPHeaders` because the header is attached cross-origin and forced CORS preflights that third parties (fonts inside the Cloudinary player iframe) rejected.
- New `tests/e2e/fixtures.js` sets both markers per worker; all specs import `test`/`expect` from it.
- `workers: 3`. Specs that mutate site-wide state (`wizard-setup`, `connection-analytics`, `deactivation-analytics`, image and video delivery) are tagged `@serial`. `npm run test:e2e` runs the parallel set, then the serial set with one worker, in the same wp-env.
- `settings-analytics` flips `image_quality` instead of `image_format` so it no longer races `media-analytics` on the same key.
- Admin API calls made with the fake e2e cloud are answered locally with a 401. `Connect::history()` issued seven ~1s real round-trips per dashboard load, which pushed `page=cloudinary` past the navigation timeout.
- `cache-analytics`: create the cache point after the admin page load, and clear the stale parent plus the asset lock before the single-purge test. `activate_parents()` deletes the unconfigured parent and leaves a 10s `_asset_lock` transient that blocks the tracked branch in `rest_purge_all()`.
- CI, e2e job: `wp-env start` runs in the background while browsers install and assets build; Playwright browsers cached by version; `Setup PHP` and a Composer cache added before `npm ci`. Without `Setup PHP` the e2e job's `npm ci` took 2.5 to 4 min on the same runs where every job with it took ~35s.
- CI, unit job: install only `@wordpress/env` (391 packages, ~15s) at the lockfile version into a scratch prefix instead of the full `npm ci`; Composer deps installed directly. The wp-env sources cache is removed: once it actually restored, it carried the previous run's install state, the plugin activation hook did not run, and PHPUnit flagged two tests risky for a missing relationships table.
- Tried and reverted: `npm ci --ignore-scripts` made the install step take 4 to 7 min with npm 10.

CI timings, latest run on this branch vs median of the 5 runs before it:

| step | before | after |
|---|---|---|
| E2E job total | 363s (327 to 391) | 302s |
| `npm ci` | 40s | 39s |
| setup (browsers, build, wp-env) | 127s | 104s |
| `Run E2E tests` | 147s | 119s (parallel 50s + serial 66s) |
| Unit job total | 187s | 121s |

Adding new specs:

- Parallel (default). Import `test`/`expect` from `./fixtures`; nothing else is needed. Use `fakeCloudinaryConnected()`, assert only on `findAnalyticsEvents( events, 'name' )` rather than the total event count, and own the settings keys the spec flips (check `rg "image_settings\[" tests/e2e` before picking one).
- Serial. Tag the describe block and say why in a comment; `npm run test:e2e` runs tagged specs in a second, single-worker pass after the parallel set:

  ```js
  // @serial: writes real credentials to `cloudinary_connect`.
  test.describe( 'Real sync flow', { tag: '@serial' }, () => {
  ```

  Tag it if the spec writes or deletes `cloudinary_connect` (any `ensureCloudinaryConnected()` or `wp cloudinary sync`), deactivates or uninstalls the plugin, or flips a global setting another spec reads without restoring it. Serial specs may still run alongside nothing else, so they are free to assume the plugin is active and connected as they left it.
- Tests within one file always run in order in both modes (`fullyParallel` is off), so per-file `beforeAll`/`afterAll` state is safe.
- Run one phase with `npm run test:e2e:parallel` or `npm run test:e2e:serial`; one file with `npx playwright test --config tests/e2e/playwright.config.js my.spec.js`.

## QA notes

- Compare the E2E job timings on this PR against a recent `develop` run (`gh run view <id> --json jobs`).
- Locally: `npm run env:start && npm run build && npm run test:e2e`. Both phases should pass. `docker exec <tests-cli> ls wp-content/uploads/` should show `analytics-capture-w0.log` through `w2.log`.
- Run `npm run test:e2e` twice in a row to confirm the cache-analytics ordering fix holds across runs.
- Open the video delivery test with `--headed` or check the Playwright trace: no CORS errors for `fonts.gstatic.com` in the console.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,6 +83,9 @@ jobs:
   unit:
     name: Unit (PHPUnit)
     runs-on: ubuntu-latest
+    env:
+      # Shell-expanded by wp-env; matches the "Cache wp-env sources" path.
+      WP_ENV_HOME: /home/runner/.wp-env
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -106,10 +109,14 @@ jobs:
       # `env:start` from re-cloning them on every run. The key covers the
       # wp-env config and version, because a change to either means different
       # sources have to be downloaded.
+      #
+      # WP_ENV_HOME is pinned at the job level because wp-env's default differs
+      # per platform (`~/.wp-env`, or `~/wp-env` on Linux runners that ship
+      # snap), and the unpinned path produced a cache that was never saved.
       - name: Cache wp-env sources
         uses: actions/cache@v4
         with:
-          path: ~/.wp-env
+          path: ${{ env.WP_ENV_HOME }}
           key: ${{ runner.os }}-wp-env-${{ hashFiles('.wp-env.json', 'package-lock.json') }}
           restore-keys: |
             ${{ runner.os }}-wp-env-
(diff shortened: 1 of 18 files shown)


gabrielcld2 (approved):
👍 Thanks for the improvements

PR #1281, opened 2026-09-15
Fix media uploaded in a non-default WPML language never finishing sync

gabrielcld2:
Fixes media uploaded from a non-default WPML language getting permanently stuck in a "syncing" state

## Approach

- When syncing an asset, the plugin queues an attachment and fires a non-blocking REST request to itself to upload it in the background.
- WPML hooks WordPress core's `home_url` filter to inject the browsing language into every generated URL. `Utils::rest_url()` builds the loopback URL by calling core `rest_url( $path )`, which applies that filter to the REST base *before* the endpoint path is appended.
  - For any non-default language this produces a malformed URL: e.g. `http://site/wp-json/?lang=fr/cloudinary/v1/queue` instead of `http://site/wp-json/cloudinary/v1/queue?lang=fr`.
- Because of this the queued attachment is left pending forever with no error anywhere.
- Fixed in `php/integrations/class-wpml.php` by hooking the existing `cloudinary_rest_url` filter:
  - Rebuild the URL with WPML's own supported `WPML_URL_Filters::remove_global_hooks()` / `add_global_hooks()` pattern (the same mechanism WPML uses internally for its Google Site Kit and canonical-URL compatibility code) to get a clean, language-unfiltered URL, wrapped in `try`/`finally` so a thrown exception can't leave WPML's URL filters disabled for the rest of the request.
  - Re-apply the language via WPML's documented `wpml_permalink` filter. `remove_global_hooks()` alone is only enough in directory/domain negotiation mode, where WPML separately hooks core's `rest_url` filter to re-insert the language after the full URL is built. In "language as a parameter" mode there's no such hook, so the clean URL would be missing `?lang=` entirely — not just for the sync loopback, but for every `Utils::rest_url()` consumer (asset fetch/save, cache purge, analytics, UI state), silently running them under the default language's context instead. `wpml_permalink` is a no-op when the URL already carries the right language (directory/domain mode) and correctly appends it when it doesn't (parameter mode).
  - Guard the undocumented `remove_global_hooks`/`add_global_hooks` methods with `method_exists()` in case a future WPML release renames or removes them.

## QA notes

- Requires a WPML install (Multilingual CMS + Media Translation) with at least one non-default language configured.
- Switch wp-admin to a non-default language (e.g. `?lang=fr`), upload a new image to the Media Library.
  - Before this fix: it stays in a "syncing" state indefinitely.
  - With this change: upload finishes syncing automatically, and the relationship row's `media_context` matches the language it was uploaded in (verified via `wp_cloudinary_relationships`).
- Sanity check the default-language path is unaffected: upload in the default language and confirm it still syncs immediately with `media_context` set to the default language.
- Tested against all three WPML language negotiation modes' URL-handling paths (directory, domain, parameter) by tracing WPML's own source; end-to-end verified in "language as a parameter" mode.


php/integrations/class-wpml.php
```diff
@@ -64,6 +64,7 @@ public function register_hooks() {
 		add_filter( 'cloudinary_media_context_query', array( $this, 'filter_media_context_query' ) );
 		add_filter( 'cloudinary_media_context_things', array( $this, 'filter_media_context_things' ) );
 		add_filter( 'cloudinary_home_url', array( $this, 'home_url' ) );
+		add_filter( 'cloudinary_rest_url', array( $this, 'rest_url' ), 10, 3 );
```
utkarshcloudinary: **Blocker: this filter is too broad.**

`cloudinary_rest_url` is the shared filter for all 15+ `Utils::rest_url()` call sites, including browser-facing endpoints: `php/class-media.php:2316` (`fetch`), `php/class-media-library.php:84`, `php/ui/component/class-asset.php:266-270` (`save_asset`, `purge_cache`), `php/ui/class-state.php:79`, `php/class-analytics.php:442`.

After this change, admin JS in a non-default language calls those endpoints on the unfiltered default-language URL. WPML resolves the request to the default language, so `add_wpml_context()` (line 160) returns the default language and `Utils::get_media_context()` returns the wrong context. Relationship rows are read and written keyed by `media_context` (`php/relate/class-relationship.php:90-107`, `php/class-delivery.php:582,660`), so an `fr` admin can create or update rows under the `en` context.

The same applies to the queue loopback itself: the REST handler now runs in the default language, so assets queued from `fr` are synced with the default `media_context`.

This also changes behavior on sites that work today. In directory mode WPML produces `http://site/fr/wp-json/cloudinary/v1/queue`, which is valid and resolves correctly; domain mode likewise. Only the parameter negotiation mode produces the malformed URL described in the PR body, but the fix applies to all three.

Suggested direction: keep the language and repair only the malformed part, for example rebuild the clean URL and then re-attach the language the way `add_locale()` already does at line 242, or limit the rebuild to the parameter negotiation mode.

php/integrations/class-wpml.php
```diff
@@ -197,6 +198,39 @@ public function home_url() {
...
+	 * @param string|null $scheme   The scheme used for the REST url.
+	 *
+	 * @return string
+	 */
+	public function rest_url( $rest_url, $path, $scheme ) {
+		if ( ! class_exists( 'WPML_URL_Filters' ) || ! function_exists( 'WPML\Container\make' ) ) {
+			return $rest_url;
+		}
+
+		$url_filters = make( 'WPML_URL_Filters' );
```
utkarshcloudinary: **Warning: `make()` can build a second `WPML_URL_Filters` instance.**

`can_enable()` only checks `ICL_SITEPRESS_VERSION`, which WPML defines at the top of `sitepress.php` (line 32 in 4.7.6). The shared instance is registered much later, by `share( [ $wpml_url_filters ] )` at line 331, after several early returns (incompatible plugin list, `WPML_UPGRADE_NOT_POSSIBLE`, multisite inactive). If WPML bailed out in between, `WPML_URL_Filters` is only registered as a shared *class*, so Auryn constructs a fresh instance. Two consequences:

- the constructor calls `add_hooks()`, which registers a duplicate set of WPML URL filters;
- `remove_global_hooks()` then removes only that new instance's callbacks, so the URL stays corrupted, and Auryn can throw `WPML\Auryn\InjectionException`, which is not caught here (other methods in this file declare `@throws` for it).

Prefer the already-shared global when it exists, and guard the container call:

```php
$url_filters = isset( $GLOBALS['wpml_url_filters'] ) ? $GLOBALS['wpml_url_filters'] : make( 'WPML_URL_Filters' );
```

plus a `try`/`catch ( InjectionException $e )` that returns `$rest_url` unchanged.

php/integrations/class-wpml.php
```diff
@@ -197,6 +198,39 @@ public function home_url() {
...
+		$url_filters->remove_global_hooks();
+		$rest_url = rest_url( $path, $scheme );
+		$url_filters->add_global_hooks();
```
utkarshcloudinary: **Warning: no `try`/`finally` around the hook removal.**

If `rest_url()` or any callback on the core `rest_url` / `home_url` filters throws, `add_global_hooks()` never runs and WPML's `home_url`, `post_link`, `post_type_link` and `get_edit_post_link` filters stay removed for the rest of the request.

```suggestion
		$url_filters->remove_global_hooks();
		try {
			$clean_rest_url = rest_url( $path, $scheme );
		} finally {
			$url_filters->add_global_hooks();
		}

		return $clean_rest_url;
```

This also avoids reassigning the `$rest_url` parameter, which hides the original value and makes a safe fallback harder to add later.

php/integrations/class-wpml.php
```diff
@@ -197,6 +198,39 @@ public function home_url() {
...
+	 * canonical-URL compatibility code for the same reason).
+	 *
+	 * @param string      $rest_url The REST url, already corrupted by WPML's `home_url` filter.
+	 * @param string      $path     The REST path that was requested.
+	 * @param string|null $scheme   The scheme used for the REST url.
+	 *
+	 * @return string
+	 */
+	public function rest_url( $rest_url, $path, $scheme ) {
+		if ( ! class_exists( 'WPML_URL_Filters' ) || ! function_exists( 'WPML\Container\make' ) ) {
```
utkarshcloudinary: Nit: consider also checking `method_exists( $url_filters, 'remove_global_hooks' )`. These methods are public but are not part of WPML's documented API, so a WPML update can remove or rename them and this would fatal rather than degrade.

php/integrations/class-wpml.php
utkarshcloudinary, line 67: **Blocker: in parameter negotiation mode this drops the language instead of repairing the URL.**

WPML hooks core's `rest_url` filter per negotiation strategy:

- Directory mode: `WPML_URL_Converter_Subdir_Strategy::convertRestUrl()` re-applies the language *after* the path is appended.
- Domain mode: `WPML_URL_Converter_Domain_Strategy::convertRestUrlToCurrentDomain()` does the equivalent for the host.
- Parameter mode: `WPML_Lang_Parameter_Filters::add_hooks()` registers nothing on `rest_url`.

`remove_global_hooks()` does not touch those strategy hooks, it only removes `home_url`, `post_link`, `post_type_link`, `attachment_link`, `wpml_filter_link`, `get_edit_post_link`, `oembed_request_post_id` and `post_embed_url`. So the inner `rest_url()` call here still gets the strategy filter applied, and directory and domain mode keep the language and produce the same URL as today.

Parameter mode is the one that breaks. Nothing puts the language back, so every `Utils::rest_url()` consumer loses `?lang=fr`, not only the background loopback. That includes browser-facing endpoints: `php/class-media.php:2316` (`fetch`), `php/class-media-library.php:84`, `php/ui/component/class-asset.php:266-270` (`save_asset`, `purge_cache`), `php/ui/class-state.php:79`, `php/class-analytics.php:442`. WPML then resolves those requests to the default language, so `add_wpml_context()` (line 160) returns the default language and `Utils::get_media_context()` returns the wrong context. Relationship rows are read and written keyed by `media_context` (`php/relate/class-relationship.php:90-107`, `php/class-delivery.php:582,660`), so an `fr` admin can create or update rows under the `en` context. The queue loopback has the same problem: the REST handler runs in the default language, so assets queued from `fr` sync with the default `media_context`.

Suggested fix: rebuild the clean URL, then re-apply the language through WPML's documented `wpml_permalink` filter, which routes to `WPML_URL_Converter::convert_url()` and places the language correctly for all three modes.

```php
public function rest_url( $rest_url, $path, $scheme ) {
	if ( ! class_exists( 'WPML_URL_Filters' ) || ! function_exists( 'WPML\Container\make' ) ) {
		return $rest_url;
	}

	$url_filters = isset( $GLOBALS['wpml_url_filters'] ) ? $GLOBALS['wpml_url_filters'] : make( 'WPML_URL_Filters' );

	$url_filters->remove_global_hooks();
	try {
		$clean_url = rest_url( $path, $scheme );
	} finally {
		$url_filters->add_global_hooks();
	}

	return apply_filters( 'wpml_permalink', $clean_url, apply_filters( 'wpml_current_language', null ) );
}
```

For `/wp-json/cloudinary/v1/queue` in parameter mode this yields `http://site/wp-json/cloudinary/v1/queue?lang=fr`, which is the URL the PR description asks for. In directory and domain mode `convert_url()` sees the language already in the URL and returns it unchanged, so nothing regresses. This file already uses the same idea at line 243 in `add_locale()`.

A smaller alternative is `add_query_arg( 'lang', apply_filters( 'wpml_current_language', null ), $clean_url )`, but that hard-codes parameter mode. `wpml_permalink` is public API since WPML 3.2.2 and covers all modes.

php/integrations/class-wpml.php
utkarshcloudinary: **Warning: no `try`/`finally` around the hook removal.**

If `rest_url()` or any callback on the core `rest_url` / `home_url` filters throws, `add_global_hooks()` never runs and WPML's `home_url`, `post_link`, `post_type_link` and `get_edit_post_link` filters stay removed for the rest of the request.

```suggestion
		$url_filters->remove_global_hooks();
		try {
			$clean_rest_url = rest_url( $path, $scheme );
		} finally {
			$url_filters->add_global_hooks();
		}

		return $clean_rest_url;
```

This also avoids reassigning the `$rest_url` parameter, which hides the original value and makes a safe fallback harder to add later.

php/integrations/class-wpml.php
utkarshcloudinary, line 217: Nit: consider also checking `method_exists( $url_filters, 'remove_global_hooks' )`. These methods are public but are not part of WPML's documented API, so a WPML update can remove or rename them and this would fatal rather than degrade.

utkarshcloudinary (commented):
Reviewed with help of AI and left feedback.

php/integrations/class-wpml.php
gabrielcld2: Good catch, thanks — verified this against the actual WPML source (`class-wpml-lang-parameter-filters.php`, `class-wpml-url-converter-subdir-strategy.php`, `class-wpml-url-converter.php`) and you're right on all counts: parameter mode has no dedicated `rest_url` hook, and `wpml_permalink` → `convert_url()` is idempotent when the language is already present, so it's safe across all three modes.

Pushed a fix in dd29364c: rebuild the clean URL as before, then `apply_filters( 'wpml_permalink', $clean_rest_url, apply_filters( 'wpml_current_language', null ) )`. Re-tested end to end — the loopback URL now comes out as `.../cloudinary/v1/queue/?lang=fr` in parameter mode, and the synced attachment's relationship row now correctly shows `media_context = fr` instead of the default.

php/integrations/class-wpml.php
gabrielcld2: Agreed — fixed in dd29364c, now wrapped in try/finally and no longer reassigns the `$rest_url` parameter.

php/integrations/class-wpml.php
gabrielcld2: Added in dd29364c — now checks `method_exists()` on both `remove_global_hooks` and `add_global_hooks` before calling them.

gabrielcld2:
@utkarshcloudinary thanks for the review! I've addressed the issues you found, and did some manual testing again to confirm the solution still works as expected. Please have another look

utkarshcloudinary (approved):
Looks good!

